PR #2, opened 2015-07-16
Accept empty clientId in connect packet generation

psorowka:
MQTT 3.1.1 allows servers to accept client connections with zero-byte
clientIds, if the cleanSession flag is set to 1.

This commit allows the generation of a connect packet with empty clientId
(`clientId: null` or `clientId: ""`). The pre-conditions `protocolVersion == 4`
(MQTT 3.1.1) and `clean == 1` are validated.

### DIFF
--- a/generate.js
+++ b/generate.js
@@ -41,7 +41,7 @@ function connect(opts) {
     , will = opts.will
     , clean = opts.clean
     , keepalive = opts.keepalive || 0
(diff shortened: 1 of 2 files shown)


psorowka:
@mcollina the build fails but not due to my changes. For me it looks as if the travis config is not up to date

PR #4, opened 2015-09-01
Removed unexpected characters (0xc2 0xa0)

pmorjan:


mcollina:
It's not really clear what's the problem and how this address that. Would you mind explaining?

mcollina:
Thanks, this one was really weird!

PR #8, opened 2016-01-14
Fixed crashing bug when parsing invalid messages

psorowka:
Fixes a crash that was observed in Mosca (https://github.com/mcollina/mosca/issues/393) when starting to parse a TLS encrypted stream while assuming to parse plain MQTT data.

So what happened? The TLS connect message that arrived at mosca looked like this:

```
<Buffer 16 03 01 01 01 01 00 00 fd 03 03 00 09 9a 85 fc 19 19 fa 06 a4
8c a5 53 5d 6e c1 01 23 be 68 00 85 b9 0c 29 8b 24 87 30 09 e6 e4 00 00
80 c0 2f c0 2b ... >
```

Note the `16` at the start, this suggests that we have a regular CONNECT packet. The next byte `3` suggests a remaining length of the packet of 3 bytes which is obviously not true, but the implementation of the parser just checks if at least 3 bytes remain in the buffer before it starts processing the payload.

The parseConnect function of the parser now checks the next two bytes `01 01` which would describe the protocol identifier string length. Usually this would be 4 or 6 bytes, but here we have 257 bytes. So the next 257 bytes are consumed into a string and our parsing pointer is at buffer position 259. Unfortunately, the buffer only has 260 bytes in total. That is why the next parsing steps produce an out of range exception..

We obviously have various problems here:

- the parser does not check for plausibility when a string ought to be longer than the packet
- the parser does not check for a valid protocol identifier string at all
- the parser does not check (at some points) if the buffer has remaining length before parsing the next bytes
- the parser does not catch any (further, unexpected) exceptions in order to safe the upper layers from running into bad behavior

All of these where fixed and respective tests added.

### DIFF
--- a/parser.js
+++ b/parser.js
@@ -36,15 +36,23 @@ Parser.prototype._newPacket = function () {
 }
 
 Parser.prototype.parse = function (buf) {
+
   this._list.append(buf)
 
-  while ((this.packet.length != -1 || this._list.length > 0) &&
-         this[this._states[this._stateCounter]]()) {
-    this._stateCounter++
+  try {
 
-    if (this._stateCounter >= this._states.length) {
-      this._stateCounter = 0
+    while ((this.packet.length != -1 || this._list.length > 0) &&
+           (noError = this[this._states[this._stateCounter]]())) {
+      this._stateCounter++
+
+      if (this._stateCounter >= this._states.length) {
+        this._stateCounter = 0
+      }
     }
+  }
+  catch (e) {
+
+    return this.emit('error', e);
   }
 
   return this._list.length
@@ -93,7 +101,7 @@ Parser.prototype._parseLength = function () {
     this._list.consume(bytes)
   }
 
-  return result
+  return result 
 }
 
 Parser.prototype._parsePayload = function () {
@@ -163,14 +171,27 @@ Parser.prototype._parseConnect = function () {
   if (protocolId === null)
     return this.emit('error', new Error('cannot parse protocol id'))
 
+  if (['MQTT', 'MQIsdp'].indexOf(protocolId) < 0) {
+
+    return this.emit('error', new Error('invalid protocol id'))
+  }
+
   packet.protocolId = protocolId
 
   // Parse constants version number
-  if(this._pos > this._list.length)
+  if(this._pos >= this._list.length)
     return this.emit('error', new Error('packet too short'))
 
   packet.protocolVersion = this._list.readUInt8(this._pos)
+
+  if([3,4].indexOf(packet.protocolVersion) < 0) {
+
+    return this.emit('error', new Error('invalid protocol version'))
+  }
+
   this._pos++
+  if(this._pos >= this._list.length)
+    return this.emit('error', new Error('packet too short'))
 
   // Parse connect flags
   flags.username  = (this._list.readUInt8(this._pos) & constants.USERNAME_MASK)
@@ -335,7 +356,7 @@ Parser.prototype._parseString = function(maybeBuffer) {
   var length = this._parseNum()
     , result
 
-  if(length === -1 || length + this._pos > this._list.length)
+  if(length === -1 || length + this._pos > this._list.length || length + this._pos > this.packet.length)
     return null
 
   result = this._list.toString('utf8', this._pos, this._pos + length)
--- a/parser.js
+++ b/parser.js
@@ -40,7 +40,7 @@ Parser.prototype.parse = function (buf) {
   this._list.append(buf)
 
   while ((this.packet.length != -1 || this._list.length > 0) &&
-         (noError = this[this._states[this._stateCounter]]())) {
+         this[this._states[this._stateCounter]]()) {
     this._stateCounter++
 
     if (this._stateCounter >= this._states.length) {
--- a/parser.js
+++ b/parser.js
@@ -39,20 +39,13 @@ Parser.prototype.parse = function (buf) {
 
   this._list.append(buf)
 
-  try {
+  while ((this.packet.length != -1 || this._list.length > 0) &&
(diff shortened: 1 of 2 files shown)


parser.js
```diff
@@ -36,16 +36,24 @@ Parser.prototype._newPacket = function () {
...
-      this._stateCounter = 0
+      if (this._stateCounter >= this._states.length) {
+        this._stateCounter = 0
+      }
     }
   }
+  catch (e) {
+
+    return this.emit('error', e);
+  }
```
mcollina: try-catch is slow, and this will really slow down everything.

Is this needed in the first place to fix this issue?

parser.js
```diff
@@ -93,7 +101,7 @@ Parser.prototype._parseLength = function () {
     this._list.consume(bytes)
   }
 
-  return result
+  return result
```
mcollina: can you please remove the space at the end?

parser.js
```diff
@@ -163,14 +171,27 @@ Parser.prototype._parseConnect = function () {
...
+  if (['MQTT', 'MQIsdp'].indexOf(protocolId) < 0) {
+
+    return this.emit('error', new Error('invalid protocol id'))
+  }
+
   packet.protocolId = protocolId
 
   // Parse constants version number
-  if(this._pos > this._list.length)
+  if(this._pos >= this._list.length)
```
mcollina: Good catch!

parser.js
```diff
@@ -163,14 +171,27 @@ Parser.prototype._parseConnect = function () {
...
     return this.emit('error', new Error('packet too short'))
 
   packet.protocolVersion = this._list.readUInt8(this._pos)
+
+  if([3,4].indexOf(packet.protocolVersion) < 0) {
+
+    return this.emit('error', new Error('invalid protocol version'))
+  }
```
mcollina: This is slow, a nice if is way better.

parser.js
psorowka: No it's not needed. All potental pitfalls i have spotted are checked for explicitely now.

Let's talk about that at another place, we should really introduce a catch *somewhere*, any bug within the processing should only kill the specific client connection and not the whole broker

parser.js
psorowka: oops

psorowka:
will push requested changes tomorrow

parser.js
```diff
@@ -163,14 +171,27 @@ Parser.prototype._parseConnect = function () {
   if (protocolId === null)
     return this.emit('error', new Error('cannot parse protocol id'))
 
+  if (['MQTT', 'MQIsdp'].indexOf(protocolId) < 0) {
```
mcollina: A nice and old `||` is faster.

parser.js
mcollina: The problem is that _any_ try catch will make things slow. Moreover, any exception in node are better to crash the process anyway, as it is hard to know if you are not leaking memory/file descriptor/whatever.

mcollina:
Also, would you mind splitting the long explanation among the various tests or complete it for the ones that are missing the explanation?

mcollina:
Anyway, this is really a __GOOD__ catch.

As soon as it lands, I'll backport it to the v3 series, so it will be catched by MQTT.js and Mosca.

psorowka:
Okay I have push everything

mcollina:
I did that myself :P. This is released as v4.0.4.

mcollina:
Also as v3.4.5.

psorowka:
By the way, out of curiosity I have added a try/catch block to the benchmark:

```
var mqtt    = require('../')
  , parser  = mqtt.parser()
  , max     = 10000000
  , i
  , start   = Date.now() / 1000
  , time

for (i = 0; i < max; i++) {
  parser.parse(new Buffer([
    48, 10, // Header
    0, 4, // Topic length
    116, 101, 115, 116, // Topic (test)
    116, 101, 115, 116 // Payload (test)
  ]))
}

console.log('No try/catch');
time = Date.now() / 1000 - start
console.log('Total packets', max)
console.log('Total time', Math.round(time * 100) / 100)
console.log('Packet/s', max / time)

start   = Date.now() / 1000

for (i = 0; i < max; i++) {
  try {
    parser.parse(new Buffer([
      48, 10, // Header
      0, 4, // Topic length
      116, 101, 115, 116, // Topic (test)
      116, 101, 115, 116 // Payload (test)
    ]))
  }
  catch (e) {

    console.log(e);
  }
}

console.log('-----------');
console.log('With try/catch');
time = Date.now() / 1000 - start
console.log('Total packets', max)
console.log('Total time', Math.round(time * 100) / 100)
console.log('Packet/s', max / time)
```

The result does not suggest a performance dip. Only thing notable is a 15% performance difference between node 0.10 and node 4

*NODE 4*
```
$ node benchmarks/parse.js 
No try/catch
Total packets 10000000
Total time 31.13
Packet/s 321192.26504089916
-----------
With try/catch
Total packets 10000000
Total time 31.28
Packet/s 319662.436545051
```

*NODE 0.10*

```
$ node benchmarks/parse.js 
No try/catch
Total packets 10000000
Total time 36.96
Packet/s 270533.4921440323
-----------
With try/catch
Total packets 10000000
Total time 36.57
Packet/s 273433.2274061726
```

mcollina:
This is actually different from putting it in the same function of a `while`.
Performance-wise, putting it in a small function that just wraps another one is ok.

psorowka:
i see.. in this case parseStream in mqtt-connection could be a candidate

psorowka:
ok just to confirm that, when putting try/catch at the original position I receive `32.79s` for node4 and `42.75s` for node 0.10. So obviously node4 is also in this respect a bit better in optimizing but still slower compared to the first benchmark.

PR #9, opened 2016-01-15
Feature/random test output

psorowka:
When a random test package triggers a unhandled parser exception, the
content of this packet as well as the stack trace are printed in order
to ease debugging

### DIFF
--- a/testRandom.js
+++ b/testRandom.js
@@ -23,8 +23,8 @@ try {
 }
 catch(e) {
 
-  console.log('Exception occured at packet ')
-  console.log(new Buffer(randomPacket))
+  console.log('Exception occured at packet')
+  console.log(randomPacket)
   console.log(e.message)
   console.log(e.stack)
 }
--- a/testRandom.js
+++ b/testRandom.js
@@ -1,12 +1,12 @@
 
 var mqtt    = require('./')
-, crypto  = require('crypto')
-, max     = 100000
-, i
-, start   = Date.now() / 1000
-, time
-, errors  = 0
-, randomPacket
+  , crypto  = require('crypto')
+  , max     = 100000
+  , i
+  , start   = Date.now() / 1000
+  , time
+  , errors  = 0
+  , randomPacket
 
 function doParse () {
   var parser  = mqtt.parser()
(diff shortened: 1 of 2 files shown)


mcollina:
Are you ok to help maintaining this module? If you are ok with CONTRIBUTING.md, I'll add you to the collaborators.

testRandom.js
```diff
@@ -6,16 +6,27 @@ var mqtt    = require('./')
...
   , time
   , errors  = 0
-
+  , randomPacket
 
 function doParse () {
   var parser  = mqtt.parser()
   parser.on('error', onError)
-  parser.parse(crypto.randomBytes(Math.floor(Math.random() * 10)))
+  randomPacket = crypto.randomBytes(Math.floor(Math.random() * 10))
```
mcollina: Possibly we should generate more longer packets, maybe up to 256 or 512 bytes.

psorowka:
Sure!

mcollina:
Feel free to add yourself to the contributors list at the end :).

testRandom.js
psorowka: I also thought we possibly should shape the stochastic distritbution of the first byte so that most packets at least enter the parser and aren't rejected by the first gate (packet type selector)

mcollina: That make sense as well :).

psorowka:
I added everything we talked about and observed a few interesting things:

- first I added a counter on the `packet` event
- when max. chunk length was set to 10, the number of chunks that caused either an `error` or a `packet` was way below 10%, assumingly because most packets just where shorter than the suggested packet length
- on the other hand when setting the max chunk length to 512 or even 1024 lead to way *more* (packets + errors) than chunks, assumingly because multiple "packets" could be found in some chunks
- this leads to the behavior that increasing chunk size enormously increases the test-time.

I would finish and merge this test now.

mcollina:
How long is now? I've set things up we run this on travis, but if that takes much time we need to short that up / make the number of rounds configurable, so in Travis is small enough.

mcollina:
All good for me, feel free to merge.

psorowka:
Duration should be fine for travis. It was more like: increasing packet length from 10 to 512 increased the time from 0.8s to 2.2s. In the first place I thought: wow, why should packet length increase the execution time? But the answer can be seen above.

PR #10, opened 2016-01-19
Fix issues when a float is used instead of a number

nightwolfz:
Some tests for MQTT.js use a float for the `keepalive` argument for mqtt.connect()

This fixes pending connections.

```
1) Websocket Client pinging should defer the next ping when sending a control packet:
     Error: timeout of 5000ms exceeded. Ensure the done() callback is being called in this test.
```

### DIFF
--- a/test.js
+++ b/test.js
@@ -754,6 +754,11 @@ testGenerateError('Invalid keepalive', {
   , password: 'password'
 })
 
+testGenerateError('Invalid keepalive', {
+    cmd: 'connect'
+    , keepalive: 3.1416
+})
+
 testGenerateError('Invalid will', {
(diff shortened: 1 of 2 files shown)


mcollina:
May I ask you to add a unit test? And also a comment in there?
Il giorno mar 19 gen 2016 alle 19:31 Ryan M. <notifications@github.com> ha
scritto:

nightwolfz:
Apologies, added a small test case and changed the code to prevent unexpected behavior (throw an error instead of rounding off floats to integers like in the previous commit). 

The performance seems unaffected.

mcollina:
Thanks!!

Issue #14, opened 2016-09-03
packet event still emitted after parse error


jdiamond:
I have a custom broker using mqtt-connection that crashed today because it received a packet that generated a "wrong subscribe header" error.

I think this is what happened:
- mqtt-packet tried to parse the packet, emitted an error event, then emitted a packet event.
- mqtt-connection forwarded the error event to me and converted the packet event into a subscribe event.
- Since the whole packet wasn't parsed, the `subscriptions` property was undefined. That's where my broker crashed.

IMO, `Parser.parse()` should skip emitting the packet event as soon as it encounters an error because it seems strange for me to have to validate the packet that mqtt-packet already determined was invalid.

Since the MQTT spec says connections should be closed on protocol violations, maybe `Parser.parse()` should stop parsing completely when it encounters an error in case there are more packets in the buffer. This would stop mqtt-connection from continuing to emit packets that shouldn't be processed.

If you agree, I can send a PR but I think this is a backwards incompatible change to the API so would require a major bump.

mcollina:
I think that is a bug, both here (it should emit more parse events) and in mqtt-connection (it should close the connection on a parse error). Would you mind to send a couple of PRs?

robertsLando:
Fixed by #15

PR #15, opened 2016-09-05
do not emit packet events and stop parsing after a parse error

jdiamond:
For #14.

### DIFF
--- a/parser.js
+++ b/parser.js
@@ -10,37 +10,35 @@ function Parser() {
     return new Parser()
   }
 
-  this._list = bl()
-  this._newPacket()
-
   this._states = [
       '_parseHeader'
     , '_parseLength'
     , '_parsePayload'
     , '_newPacket'
   ]
-  this._stateCounter = 0
+
+  this._resetState()
 }
 
 inherits(Parser, EE)
 
-Parser.prototype._newPacket = function () {
-  if (this.packet) {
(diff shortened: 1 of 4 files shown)


parser.js
```diff
@@ -10,37 +10,35 @@ function Parser() {
...
-  return true
+  this.error = null
+  this._list = bl()
+  this._stateCounter = 0
 }
 
 Parser.prototype.parse = function (buf) {
+  if (this.error) {
+    this._resetState()
+  }
```
mcollina: I think we should return early in the `if`, otherwise it keep resetting itself: if we want to parse two `buffer`, and the first errors, the second should be ignored. Or maybe errored, what do you think?

mcollina:
Would you mind updating `.travis.yml` here as well?

parser.js
jdiamond: I could go either way. `_resetState()` resets the error as well as the buffer that had the error in it so it shouldn't keep resetting.

If we don't reset the state, calling `parse()` again with a new buffer is guaranteed to keep emitting the same error over and over since it appends the new buffer after the old.

So we could just tell the user to throw that instance of `Parser` away, or reset the internal state like I'm doing here. If we do choose not to reset for them, should we at least throw an exception to make it obvious the current instance is broken?

PR #16, opened 2016-09-09
return on first error to avoid writing invalid bytes

jdiamond:
I added `return` to all the lines that emit error events. This stops writing bytes to the output stream that we know are invalid.


mcollina:
Thanks @jdiamond! We should return either `true` or `false`. `stream.emit('error')` returns the streams, which is truthy, and that might cause issues. I think we should return `false`.

mcollina:
Perfect, thanks!

PR #20, opened 2016-10-11
Update parser.js

nitinbansal1989:


mcollina:
Why?
Can you add a unit test?

nitinbansal1989:
I ran a load test from MQTTBOX client and it was sending reserved type of messages.
I checked it with mosquitto server and it was working fine with "reserved" type packets. 
So i thought it might be helpful if the this packet type is supported otherwise throwing exception.

mcollina:
Ok, can you add a unit test?

nitinbansal1989:
i think it would be something like this. Please review. Sry but i don't know the exact header value.

testParseGenerate('reserved', {
  cmd: 'reserved',
  retain: false,
  qos: 0,
  dup: false,
  length: 0
}, new Buffer([
  224, 0 // Header
]))

mcollina:
can you please add it as a commit? Thanks.

nitinbansal1989:
Test case added. Can you publish the updated package.

mcollina:
the test is not part of this PR.

mcollina:
On second thought.. why is MQTTBOX sending "reserved" packets?
Currently it errors the parser: https://github.com/mqttjs/mqtt-packet/blob/master/parser.js#L130.

This is a semver-major change if we add it, and I am not 100% sure it's a good idea.

I do not see https://github.com/workswithweb/MQTTBox the reserved packet being used.

nitinbansal1989:
I also don't know, but it is sending them during load testing.

mcollina:
It seems some data is getting corrupted during load testing. What are you benchmarking?

nitinbansal1989:
I running the load testing on mosca server.

mcollina:
Which version?
I'm not familiar with MQTTBox, is it working over websocket?

nitinbansal1989:
You can add chrome extension of MQTT Box and it features as a standalone mqtt client app.
I check and found the following "reserved" packet header in mqtt specs.
http://public.dhe.ibm.com/software/dw/webservices/ws-mqtt/mqtt-v3r1.html#fixed-header
I think that it should be added if specified in specs.

mcollina:
I do not think so. That packet is not supported and it should not be using.
If MQTTBox is sending it, it's their fault.
Receiving a reserved packet errors the parser, and I think that is the right behavior.

IMHO this happens because some bits become corrupted or broken within the stream.

nitinbansal1989:
Ok. Can you suggest me some other client where i can do a load test or something.

mcollina:
This should work correctly: https://github.com/mqttjs/MQTT.js/blob/master/benchmarks/bombing.js

wuhkuh:
@mcollina is correct, if we follow the MQTT 3.1.1 [specification](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Table_2.1_-) reserved packets may not be handled.

nitinbansal1989:
@mcollina Thanks

PR #22, opened 2016-12-17
tweak readme

ralphtheninja:


README.md
```diff
@@ -240,12 +240,12 @@ All properties are mandatory.
 }
 ```
 
-Only the `topic` and properties are mandatory
+Only the `topic` and `payload` properties are mandatory.
```
mcollina: Are you sure about this? An empty payload should be possible.

README.md
```diff
@@ -240,12 +240,12 @@ All properties are mandatory.
...
 Both `topic` and `payload` can be `Buffer` objects instead of strings.
 `messageId` is mandatory for `qos > 0`.
 
 If `payload` is passed to `generate(packet)` as a string, it will be
-automatically converted into a `Buffer`.
+automatically converted to a `Buffer`.
```
mcollina: This might not be true. If writeToStream is used, the payload is written as a string if it is a string.

README.md
ralphtheninja: Nope, I'm not sure about this. I just found the sentence "Only the `topic` and properties are mandatory" a bit odd :)

ralphtheninja: Maybe change it to "Only the `topic` property is mandatory"?

ralphtheninja: Yes you're right. If omitted it will default to `empty`

https://github.com/mqttjs/mqtt-packet/blob/master/writeToStream.js#L221

README.md
```diff
@@ -240,13 +240,10 @@ All properties are mandatory.
...
 ```
 
-Only the `topic` and properties are mandatory
+Only the `topic` property is mandatory.
 Both `topic` and `payload` can be `Buffer` objects instead of strings.
 `messageId` is mandatory for `qos > 0`.
 
-If `payload` is passed to `generate(packet)` as a string, it will be
-automatically converted into a `Buffer`.
-
```
ralphtheninja: See https://github.com/mqttjs/mqtt-packet/blob/master/writeToStream.js#L258 and https://github.com/mqttjs/mqtt-packet/blob/master/generate.js#L21

ralphtheninja:
Once we get this correct we should probably look over aedes docs as well https://github.com/mcollina/aedes/pull/78 (which I updated partly based on this documentation)

mcollina:
Thanks, this is perfect. Would you mind updating the ones at Aedes as well?

ralphtheninja:
> Thanks, this is perfect. Would you mind updating the ones at Aedes as well?

Will do!

Issue #28, opened 2017-05-09
Add ability to configure numCache


kirilknysh:
In current implementation a `numCache` is create by default (file numbers.js) and it contains 65536 buffers with predefined values. That leads to quite high memory consumption: ~14Mb of memory (node process) can be saved.

For sure it's always a trade off between memory and performance that's why I'd like to propose to configure `numCache`: to give an ability to either create cache during start or create buffers on demand and, potentially, cache created.

I did a quick comparison (on `benchmarks/writeToStream.js`):

- current implementation gives
```
Total time 2703
Total packets 1000000
Packet/s 369959.3044765076
```

- a quickly patched version:
```
var buffer;

const handler = {
    get: function (target, name) {
        const i = ~~name;

        buffer = new Buffer(2);
        buffer.writeUInt8(i >> 8, 0, true);
        buffer.writeUInt8(i & 0x00FF, 0 + 1, true);
        return buffer;
    }
};

module.exports = new Proxy({}, handler);
```
gives:
```
Total time 3156
Total packets 1000000
Packet/s 316856.7807351077
```

Of course this is not a "production"-ready patched code and it can be improved. That's rather an example to show that performance degradation is not so big.

I do understand that most developers prefer speed over memory consumption that's why I propose to leave it as an option.

Please let me know if it works for you. I can prepare a PR with a good solution.

Thanks!

mcollina:
I would not merge any PRs that have any form of performance degradation.

How did you measured the 14MB? 

2 * 65536  = 131072 _bytes_ = 128 KB.

Something else should be going on. Maybe we need to allocate a big buffer, and then slice through it.

kirilknysh:
I just took the existing `number.js`, added `setInterval(() => {}, 9999);` (just not to finish the process immediately) and run `node number.js`. Then I measured memory with just regular `top` command (MacOS Sierra, 10.12.4, node v6.5.0).

Node process with the original `number.js` - 22Mb
Node process with patched code - 9Mb

Keep in mind that this is not JS heap but node process itself.

mcollina:
@kirilknysh so, how do you think we can make this configurable? Also, we might check if using an array improve things.

kirilknysh:
I've tested with an array. Regarding memory - no big difference in both: memory usage and performance.

[...]

PR #29, opened 2017-05-11
Add option to enable/disable numbers cache

kirilknysh:
Fixes https://github.com/mqttjs/mqtt-packet/issues/28

Add a flag `cacheNumbers` to control the numbers cache. Disabling the cache may save up to 14Mb of memory (node process).

Performance (`benchmarks/writeToStream.js`, average of 10 runs, Packet/s):
- before changes:                   386675.1239634539
- after changes (w/ cache):   376886.5427892937
- after changes (w/o cache): 353188.7654481942

### DIFF
--- a/writeToStream.js
+++ b/writeToStream.js
@@ -41,6 +41,10 @@ function generate (packet, stream) {
       return false
   }
 }
+/**
+ * Controls numbers cache.
+ * Set to "false" to allocate buffers on-the-flight instead of pre-generated cache
+ */
 generate.cacheNumbers = true
 
 function uncork (stream) {
--- a/writeToStream.js
+++ b/writeToStream.js
@@ -7,6 +7,8 @@ var zeroBuf = Buffer.from([0])
 var numCache = require('./numbers')
 var nextTick = require('process-nextick-args')
 
+var getNumber = numCache.getCachedNumber
+
 function generate (packet, stream) {
   if (stream.cork) {
     stream.cork()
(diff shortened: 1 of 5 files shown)


writeToStream.js
```diff
@@ -41,6 +41,7 @@ function generate (packet, stream) {
       return false
   }
 }
+generate.cacheNumbers = true
```
mcollina: I think we can use a setter here, and if it is setted to false, clear `numCache`.

writeToStream.js
```diff
@@ -520,7 +521,7 @@ function writeString (stream, string) {
  * @api private
  */
 function writeNumber (stream, number) {
-  return stream.write(numCache[number])
+  return stream.write(numCache.get(number, generate.cacheNumbers))
 }
```
mcollina: I think we can have two functions, `writeNumberCached` and  `writeNumberAllocate`, and we can assign them to `writeNumber` in the `cacheNumbers` setter.

writeToStream.js
kirilknysh: Then it would require:
either load and generate cache and then potentially clean it
or set `cacheNumbers` always (either true or false) to trigger the selection of required `writeNumber`.

mcollina: You can populate the cache when `writeToStream` is called.

kirilknysh:
- after changes w/ cache: 362711.861420035
- after changes w/o cache: 328807.3007557509

Memory consumption is the same as in the PR description

writeToStream.js
```diff
@@ -520,7 +535,7 @@ function writeString (stream, string) {
  * @api private
  */
 function writeNumber (stream, number) {
-  return stream.write(numCache[number])
+  return stream.write(getNumber(number))
 }
```
mcollina: I think we can replace that function, and have `numCache` ported here. We should really avoid that `getNumber()` function call.

kirilknysh: Not sure I understand you correctly. We anyway need a function call in case of buffer allocation (w/o cache). So `getNumber` is a function only for common interface of getting buffers.

mcollina: I've added the change that I had in mind, have a look!

Also, can you add some unit tests for this?

kirilknysh: Sure, will add

mcollina:
Can you add a unit test?

kirilknysh:
Yes, it's in progress. Sorry, takes more time than expected

PR #33, opened 2018-01-22
add support to empty string username and password

ophentis:
This adds support of username being empty string and password being empty buffer. 
In the packet, flag will be set, length will be 0 to either or both.

### DIFF
--- a/writeToStream.js
+++ b/writeToStream.js
@@ -165,7 +165,7 @@ function connect (opts, stream) {
   // Username
   var providedUsername = false
   if (username != null) {
-    if (typeof username === 'string') {
+    if (isStringOrBuffer(username)) {
       providedUsername = true
       length += Buffer.byteLength(username) + 2
     } else {
@@ -181,7 +181,7 @@ function connect (opts, stream) {
       return false
     }
 
-    if (typeof password === 'string' || Buffer.isBuffer(password)) {
+    if (isStringOrBuffer(password)) {
       length += byteLength(password) + 2
     } else {
       stream.emit('error', new Error('Invalid password'))
@@ -226,10 +226,10 @@ function connect (opts, stream) {
 
   // Username and password
   if (username != null) {
-    writeString(stream, username)
+    writeStringOrBuffer(stream, username)
   }
   if (password != null) {
-    writeStringOrBuffer(stream, Buffer.from(password))
+    writeStringOrBuffer(stream, password)
   }
   // This is a small packet that happens only once on a stream
   // We assume the stream is always free to receive more data after this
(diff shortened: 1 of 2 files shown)


mcollina:
CI is failing on node 8.

ophentis:
yup, I'll take a look.

ophentis:
Running the same flow on same version is a pass on my local machine. Something fishy here.

ophentis:
The problem I assume would happen random on all version. Luckily it's caught on one of them.

writeToStream.js
```diff
@@ -173,8 +175,8 @@ function connect (opts, stream) {
   }
 
   // Password
-  if (password) {
-    if (password.length) {
+  if (password != null) {
+    if (providedUsername && (typeof password === 'string' || Buffer.isBuffer(password))) {
```
mcollina: I think it's possible to have a password without a username. So maybe this check should be relaxed. Can you check in the spec?
Even if it is not, it will be emitting "invalid password", which is not correct: it should be throwing something like "a username is required to use a password".

ophentis: From [v3.1.1](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html)

* [MQTT-3.1.2-20] If the Password Flag is set to 0, a password MUST NOT be present in the payload.
* [MQTT-3.1.2-21] If the Password Flag is set to 1, a password MUST be present in the payload.
* [MQTT-3.1.2-22] If the User Name Flag is set to 0, the Password Flag MUST be set to 0.

ophentis: I'll work on the error.

mcollina: Fantastic

writeToStream.js
```diff
@@ -218,9 +225,12 @@ function connect (opts, stream) {
   }
 
   // Username and password
-  if (username) writeStringOrBuffer(stream, username)
-  if (password) writeStringOrBuffer(stream, password)
-
+  if (username != null) {
+    writeString(stream, username)
+  }
```
mcollina: can you restore this? We supported a Buffer before.

writeToStream.js
```diff
@@ -577,4 +588,8 @@ function byteLength (bufOrString) {
   else return Buffer.byteLength(bufOrString)
 }
 
+function isStringOrBuffer (field) {
+  return typeof field === 'string' || Buffer.isBuffer(field)
```
mcollina: you can probably check `field instanceof Buffer` here. That's what `Buffer.isBuffer()`  is doing anyway.

ophentis: Yes, but I actually want to use util.isString to check string type to make it more functional. But deprecated function stops me.
Thanks for all the time reviewing.

PR #37, opened 2018-03-10
Remove `noAssert` from buffer calls.

BridgeAR:
Refs https://github.com/nodejs/node/pull/18395

Ping @mcollina

### DIFF
--- a/numbers.js
+++ b/numbers.js
@@ -6,8 +6,8 @@ var cache = {}
 
 function generateBuffer (i) {
   var buffer = Buffer.allocUnsafe(2)
-  buffer.writeUInt8(i >> 8, 0, true)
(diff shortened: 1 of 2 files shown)


mcollina:
This guarantees increased performance in Node 4, 6, and 8. Why removing it now?

BridgeAR:
That is your call. I personally do not believe that this is going to have any measurable performance difference when being used.

I went ahead and checked for all such entries to find modules that break because of the change. That was the case a couple of times (but not here).

Feel free to close this, if you think it should stay in there.

mcollina:
It did when I put this in during the crankshaft era. I’ll check the benchmarks.

mcollina:
performance is identical.

PR #38, opened 2018-04-17
Fix inverse logic when calculating payload length during publish

falkTX:
I was getting weird errors in my application that uses mqtt and protobuf.
After some debugging, it turns out the size of the message I send is not correct.
Going through the code step by step I noticed the point of failure being mqtt-packet publish function.

You're using `Buffer.byteLength` when the payload is *not* a Buffer.
I believe this to be a typo. My application now works properly with the fix.

### DIFF
--- a/writeToStream.js
+++ b/writeToStream.js
@@ -272,7 +272,7 @@ function publish (opts, stream) {
   }
 
   // Get the payload length
-  if (!Buffer.isBuffer(payload)) length += Buffer.byteLength(payload)
+  if (Buffer.isBuffer(payload)) length += Buffer.byteLength(payload)
   else length += payload.length
 
   // Message ID must a number if qos > 0


mcollina:
Can you please add a unit test?

writeToStream.js
```diff
@@ -272,7 +272,7 @@ function publish (opts, stream) {
   }
 
   // Get the payload length
-  if (!Buffer.isBuffer(payload)) length += Buffer.byteLength(payload)
+  if (Buffer.isBuffer(payload)) length += Buffer.byteLength(payload)
```
mcollina: we need to call this for `Buffer.isBuffer(str)`? `str`  could be `'utf-8'`.

falkTX:
Unit test is tricky, since the issue does not appear on all browsers.
I am also new to javascript/node development, so some assistance for this would be nice.

writeToStream.js
falkTX: Perhaps checking if type is string first?
(My use case always uses binaries.)

mcollina: What do you mean by binary?

falkTX: It's protobuf, so a buffer/array of "random" values that cannot be converted into strings

mcollina: > It's protobuf, so a buffer/array of "random" values that cannot be converted into strings

What type is that?

mcollina:
I think the best approach is to remove the if altogether and call `+= Buffer.byteLength(payload)` in any case.

writeToStream.js
falkTX: It's a Uint8Array

PR #39, opened 2018-05-06
Adding Typescript typings.

saboya:
Packet typings were basically taken from https://github.com/mqttjs/MQTT.js/blob/master/types/lib/types.d.ts, with a minor addition: the ISubscription interface. This is due to me trying to write Aedes typescript bindings, and that additional interface would be helpful there.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "5.5.0",
   "description": "Parse and generate MQTT packets like a breeze",
   "main": "mqtt.js",
+  "types": "types/index.d.ts",
   "contributors": [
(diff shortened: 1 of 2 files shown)


mcollina:
Would these be compatible with the ones in https://github.com/mqttjs/MQTT.js/blob/master/types/lib/types.d.ts? I mean, if we add these here should we change them in MQTT.js to use this instead? Can you prepare that PR as well?

saboya:
@mcollina PR https://github.com/mqttjs/MQTT.js/pull/819 is ready.

mcollina:
Thanks!

PR #41, opened 2018-05-30
Feature mqtt 5

scarry1992:
Hi. We've implemented the MQTT 5.0 support. Please have a look and merge if it's OK. We will also make similar pull-requests to mqtt-connection and MQTTJS repos.


scarry1992:
HI, thanks for approving my commit. I would like to help maintaining all libraries as free time. I think, we should release all libraries as major version, despite the fact that the api is not breaking.

mcollina:
Feel free to send a PR adding yourself in the README and package.json of this module.

PR #46, opened 2018-12-03
updated: newest widely supported ecmascript spec

RyLeeHarrison:


parser.js
```diff
@@ -1,626 +1,648 @@
...
-var Packet = require('./packet')
-var constants = require('./constants')
+const bl = require('bl')
+const { EventEmitter } = require('events')
+const Packet = require('./packet')
+const constants = require('./constants')
 
-function Parser (opt) {
-  if (!(this instanceof Parser)) return new Parser(opt)
+class Parser extends EventEmitter {
```
mcollina: This change is actually changing behavior. Functionally its different because the previous code allowed `Parser(opt)`  to work.

PR #49, opened 2019-02-16
Update constants.js

mukeshydv:
Made server keep alive property 2-byte integer

Reference: 
https://docs.oasis-open.org/mqtt/mqtt/v5.0/cos01/mqtt-v5.0-cos01.html#_Toc514847901
https://docs.oasis-open.org/mqtt/mqtt/v5.0/mqtt-v5.0.pdf


mukeshydv:
> is this change needed also for MQTT 3.1.1?

This is related to properties, which is added in 5 so I don't think so.

mcollina:
@scarry1992 can you please check this one?

PR #60, opened 2019-06-26
Some properties can be multiple

scarry1992:
For example: https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901117


scarry1992:
@mcollina, review pls and release in npm with path version.

scarry1992:
@mcollina, I added new info in docs. Check pls.

PR #67, opened 2019-09-20
get ArrayBuffer payload length NaN

tomcat008:
When I use MQTT.js in wechat and publish ArrayBuffer message，the subclient cannot recv the message.
I debugged the program and found get the ArrayBuffer payload length "payload.length" is undefined.
And in _stream_writable.js "write" method the payload encoding is converted to “utf-8”,not "buffer"。
I've made changes and passed the test in the project.


mcollina:
Would you mind to add a unit test?

PR #69, opened 2019-10-24
Confirm should not have a zero property length

joshperry:
cref rfc 3.4.2.2.1


joshperry:
I ran into this with mosquitto, but I'm not 100% sure it's compliant with all MQTT5 implementations. When sending a message to mosquitto that has no subscribers, it acks with reason 16 and a remaining length of 3.

PR #70, opened 2019-10-24
Fix return value of unsubscribe packet

pachirel:
This method returned `undefined` even though others returned truthy value.
(`writeString` method doesn't have return value.)

Depending on how it was used, there was an issue in stream processing in my application.

<s>P.S. I'm not sure which is the correct return value type of `generate` method:`boolean` or `number`. I chose boolean followed by other methods.</s>

Thanks!

(edited): 

The return value is used in `mqtt-connection/lib/writeToStream.js`.
https://github.com/mqttjs/mqtt-connection/blob/master/lib/writeToStream.js#L13

I guess this method expects a return value of `writeToStream()` is a return value of `stream.write()`

### DIFF
--- a/test.js
+++ b/test.js
@@ -60,6 +60,16 @@ function testParseGenerate (name, object, buffer, opts) {
 
     t.equal(parser.parse(fixture), 0, 'remaining bytes')
   })
+
+  test(name + ' writeToStream', function (t) {
+    var stream = WS()
+    stream.write = () => true
+    stream.on('error', (err) => t.fail(err))
+
+    var result = mqtt.writeToStream(object, stream, opts)
+    t.equal(result, true, 'should return true')
+    t.end()
+  })
 }
 
 function testParseError (expected, fixture, opts) {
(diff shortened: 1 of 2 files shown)


mcollina:
Would you mind adding a unit test?

Il giorno ven 25 ott 2019 alle 01:49 Yuta Shimizu <notifications@github.com>
ha scritto:

pachirel:
> Would you mind adding a unit test? 

I pushed a unit test 👍 I confirmed that it has been failed without 2f1ad3a.
Could you please review my test code?

pachirel:
I have another idea to fix this issue: make `writeString()` return a result of `stream.write()`. 
I'll follow your choice.

pachirel:
@mcollina Hi, gentle reminder as I'd like to hear your feedback. :) Thanks for your attention.

mcollina:
Sorry for the delay @pachirel! I think your recommendation of making `writeString()` return the result of `stream.write()` is better.

pachirel:
@mcollina thank you for the reply! I applied your feedback.

Issue #68, opened 2019-09-24
generate function cannot handle zero-length client identifiers


janniswarnat:
Hi all,

thank you for providing this very helpful package :-).

The MQTT standard version 5 allows clients to connect using a zero-length client identifier. Yet the generate function seems not to be able to handle this:

```
const mqtt = require('mqtt-packet');
const opts = {protocolVersion: 5};
const parser = mqtt.parser(opts);

parser.on('packet', (packet) => {
    console.log("packet event emitted on noopParser", packet);
});

parser.on('error', (error) => {
    console.log("error event emitted on noopParser", error.message);
});

let packet = {
    cmd: "connect",
    protocolId: "MQTT",
    protocolVersion: 5,
    clean: true,
    keepalive: 60,
    properties:
        {
            receiveMaximum: 20
        },
    clientId:""
};

let generatedBuffer = mqtt.generate(packet, opts);
console.log('generatedBuffer',generatedBuffer);
parser.parse(generatedBuffer);
```

This will give error  message "Packet too short" when parsing the generated buffer.

Thanks in advance for your feedback and best regards

Jannis

PR #71, opened 2019-11-19
Fix issue #68: Allow zero-length client identifiers also for MQTT version 5

janniswarnat:
Allow zero-length client identifiers also for MQTT version 5

### DIFF
--- a/writeToStream.js
+++ b/writeToStream.js
@@ -111,9 +111,9 @@ function connect (packet, stream, opts) {
     return false
   } else length += 1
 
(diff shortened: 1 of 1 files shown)


mcollina:
I'm sorry I just saw this. Would you mind to add a unit test?

janniswarnat:
Can you take the code snippet from #68? Or would you like me to add something similar to test.js?

mcollina:
Please include a unit test in `test.js`.

PR #74, opened 2020-02-13
chore: add debug logs

YoDaMa:


mcollina:
CI is failing due to some extra semicolons, would you mind to fix?

YoDaMa:
@mcollina could you merge

mcollina:
This is missing to add the `debug` as a dependency. It probably work here because it's installed by some other package, but that's definitely not a given. It would also not work with yarn pnp or pnpm.

YoDaMa:
@mcollina got it. Fixed that bug. Let me know if you could merge now

YoDaMa:
@mcollina :D

YardWill:
This is a good feature, but one thing i want to say is that it will lead the package to throw Error when in IE11. Because there has many arrow function in the 'debug' package. The debug package author didn't update the main file to dist file in many issues.  Can we remove debug in the dist file? Just use it in dev mode.

PR #75, opened 2020-03-03
fix mqtt5 properties write error handling

scarry1992:
I have fixed behavior on try to write wrong value in int-type properies by MQTT 5


scarry1992:
@mcollina, review pls and merge, its critical as I see.

scarry1992:
@mcollina, and need to publish to npm, thx

PR #79, opened 2020-04-20
fix clientid bug

linzhennan:
support clientId with utf8


mcollina:
Can you please add a unit test?

PR #85, opened 2020-08-24
Updated code to ES6

seriousme:
I have updated all code to ES6 in two commits:

- The first commit updates tests to ES6 to see if the tests would break in CI with the module still using pre-ES6 code
- The second commit updates the rest of the module to ES6 including the benchmarks and code examples in the README.

All code, including examples in the README, has been converted using the [ES5 to ES6 extension of Visual code](https://marketplace.visualstudio.com/items?itemName=jeremyrajan.vscode-lebab) and then checked by `standard --fix`.
(the code examples in the README were extracted to tempfiles, converted and pasted back)
I only did manual updates to code to make `standard --fix` succeed. 

This process leaves room for further optimizations (e.g. creating javascript classes, removing some dependencies). I'm happy to try those as well but I thought I'd first do a most basic conversion and work from that.

In the README I also updated the Node versions tested in CI ;-)

Kind regards,
Hans

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Encode and Decode MQTT 3.1.1, 5.0 packets the node way.
   * <a href="#contributing">Contributing</a>
   * <a href="#license">License &amp; copyright</a>
 
-This library is tested with node v4, v6 and v7. The last version to support
+This library is tested with node v6, v8, v10, v12 and v14. The last version to support
 older versions of node was mqtt-packet@4.1.2.
 
 Installation
@@ -28,8 +28,8 @@ Examples
 ### Generating
 
 ```js
-var mqtt = require('mqtt-packet')
-var object = {
+const mqtt = require('mqtt-packet');
+const object = {
   cmd: 'publish',
   retain: false,
   qos: 0,
@@ -37,8 +37,8 @@ var object = {
   length: 10,
   topic: 'test',
   payload: 'test' // Can also be a Buffer
-}
-var opts = { protocolVersion: 4 } // default is 4. Usually, opts is a connect packet
+};
+const opts = { protocolVersion: 4 }; // default is 4. Usually, opts is a connect packet
 
 console.log(mqtt.generate(object))
 // Prints:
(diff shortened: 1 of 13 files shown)


mcollina (commented):
can you please check the benchmarks if there is a regression?

seriousme:
Well, there are lies, darn lies and benchmarks :-)

Lets look at CI results first: comparing the [previous CI results](https://github.com/mqttjs/mqtt-packet/actions/runs/221750796) listed as ES5 to the [CI results for this PR](https://github.com/mqttjs/mqtt-packet/actions/runs/222515402) listed as ES6


  | Time in seconds |   |  
-- | -- | -- | --
Node version | ES5 | ES6 | Difference |  
6 | 2,51 | 2,61 | 3,98% |  
8 | 2,01 | 2,31 | 14,93% |  
10 | 1,19 | 1,1 | -7,56% |  
12 | 1,13 | 1,34 | 18,58% |  
14 | 1,02 | 1,24 | 21,57% |  



From this comparison we would be stupid to migrate from ES5 to ES6, but remember this is cloud and VM performance can vary quite some and Github does not promise any performance on its Actions :-(

It also does not make a lot of sense that the V8 team would be improving their performance for ES5 but not for ES6 nor that v10 would be way faster than v14 (the module is not that exotic imho).

So I redid the tests on my local machine (linux container on x86 chromebook). There are probably processes running in the background as well but the results are more stable and logical than from Github Actions:

  | Time in seconds |   |  
-- | -- | -- | --
Node version | ES5 | ES6 | Difference |  
6 | 7,19 | 7,2 | 0,14% |  
8 | 6,11 | 6,11 | 0,00% |  
10 | 4,54 | 4,69 | 3,30% |  
12 | 4,76 | 4,87 | 2,31% |  
14 | 4,91 | 4,84 | -1,43% |  

Here it shows that the difference in performance is marginal as to be expected.

Running the same npm ci for ES6 on v14 5 times results in values between 4.82 and 4.97.

So my conclusion would be that the performance difference is minimal.
(and that benchmark results from testing on Github  Actions should be handled with care ;-))

Kind regards,
Hans

mcollina (approved):
lgtm

seriousme:
FYI: This PR is a more recent version of #46 which can now also be closed imho.

PR #89, opened 2020-09-03
Added extra tests

seriousme:
Hi,

I did some coverage testing and from that added some extra test helper functions and extra tests

Also:
- added an optional name parameter to `testGenerateError` so we can see the difference between different tests that have the same error message.
- modified the `connect all strings generate` to use the new `testGenerateOnly`
- moved the `unknow command` test up to sort/group the tests a bit
- replaced `MQTT5 ` in test descriptions by  `MQTT 5 ` so its easier do find all MQTT 5 stuff using grep

Kind regards,
Hans

### DIFF
--- a/test.js
+++ b/test.js
@@ -69,6 +69,15 @@ function testParseGenerate (name, object, buffer, opts) {
   })
 }
 
+// the API allows to pass strings as buffers to writeToStream and generate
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

PR #90, opened 2020-09-03
Fix: restrict Variable Byte Integer to 24bits

seriousme:
The MQTT 5.0 specification defines [Variable Byte Integer](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901011)  as a maximum of 24 bits. The 3.x  spec uses this format for packet length only but in 5.0 it is also used in properties (e.g. [subscriptionIdentifier](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901117))

Testing _generation_ of a Variable Byte Integer with anything bigger than that resulted in a  `Attempt to write outside buffer bounds` error because numbers.js would allocate 0 bytes and still try to write 4 or more bytes.
_Parsing_ of a Variable Byte Integer (e.g. in packet length) > 24 bits would not emit an error where it should.

This PR fixes that and adds tests to confirm 24 bits maximum.

It also renames the constants and variables used by the parser so they are inline with its intended use again.

Performance should not be impacted because:
a)  generator algorithm has been simplified so it should be a tiny bit faster
b)  the cache hides the performance of generation anyway :-)
c)  the rest of the changes are only small logic updates

Kind regards,
Hans

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -37,8 +37,8 @@ protocol.QOS_SHIFT = 1
 protocol.RETAIN_MASK = 0x01
 
 /* Length */
-protocol.LENGTH_MASK = 0x7F
-protocol.LENGTH_FIN_MASK = 0x80
+protocol.VARBYTEINT_MASK = 0x7F
+protocol.VARBYTEINT_FIN_MASK = 0x80
 
 /* Connack */
 protocol.SESSIONPRESENT_MASK = 0x01
--- a/numbers.js
+++ b/numbers.js
@@ -1,6 +1,13 @@
 const max = 65536
 const cache = {}
 
+// in node 6 Buffer.subarray returns a Uint8Array instead of a Buffer
+// later versions return a Buffer
+// alternative is Buffer.slice but that creates a new buffer
+// creating new buffers takes time
+// SubOk is only false on node < 8
+const SubOk = Buffer.isBuffer(Buffer.from([1, 2]).subarray(0, 1))
+
 function generateBuffer (i) {
   const buffer = Buffer.allocUnsafe(2)
   buffer.writeUInt8(i >> 8, 0)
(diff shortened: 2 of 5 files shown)


mcollina:
Test fails on node 6.

Given we use the same logic for MQTT 3.1.1, would there be compatibility problems?

seriousme:
Turns out that in Node 6  handles Buffer.subarray different from later versions 

Node 6:
```javascript
> b= Buffer.from([1,2,3,4])
<Buffer 01 02 03 04>
> b.subarray(0,1)
Uint8Array [ 1 ]
```
Node 8 and later
```javascript
> b= Buffer.from([1,2,3,4])
<Buffer 01 02 03 04>
> b.subarray(0,1)
<Buffer 01>
```

Buffer.slice works as well but creates a copy that needs to be allocated and thus costs some performance :-(
Solved now.

Kind regards,
Hans

seriousme:
As to compatibility :

MQTT [3.1](https://public.dhe.ibm.com/software/dw/webservices/ws-mqtt/mqtt-v3r1.html#msg-format),  [3.11](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718023) and [5.0](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901011) all use the same algorithm with the same 24 bits max (decimal 268,435,455, hex 0x0f ff ff ff) but only use it for length and do not name it Variable Byte Integer yet.  So no compatibility impact :-)

Kind regards,
Hans

mcollina (approved):
lgtm

Issue #95, opened 2020-11-05
Regression in v6.6.0 for parsing multi-byte length header that is split across write calls


twegener-embertec:
The following commit altered parser.js _parseLength such that it generates an error state whenever _parseVarByteNum returns false:

```
commit 3eb494a19d6a5c855ab098ab6975a934b0bff00a
Author: Hans Klunder <hans.klunder@gmail.com>
Date:   Tue Sep 8 11:59:32 2020 +0200

    Fix: restrict Variable Byte Integer to 24bits (#90)
```

That was to cater for the new max bytes check in _parseVarByteNum, that breaks the case where the remainder of the length header bytes are not available yet (this._list.length <= bytes), since that case is not an error, since it is just waiting for more bytes to be available.

It is relatively rare that it hits this problem, as it only happens if delayed incoming bytes occur at just the right (wrong) time, i.e. when parsing a multi-byte length header field. However, under high traffic we see this occurring a couple times a day since upgrading to mqtt-packet 6.6.0. I have tested winding back to 6.5.0 and we don't hit the problem there.

This is quite serious as when this occurs the byte stream does not get consumed properly, and after the first "Error: Invalid length" error occurs, parsing subsequent incoming bytes gets totally confused and emits numerous errors (losing the data), until the mqtt connection is dropped and re-established.

twegener-embertec:
Here are some unit tests that demonstrate the problem (the first for a sanity check for the test approach, then the second revealing the bug). These tests pass on v6.5.0, but the latter test fails with v6.6.0:

```
test('split length publish longer', t => {
  t.plan(3)

  const length = 255
  const topic = 'test'
  const payloadLength = length - topic.length - 2

  const parser = mqtt.parser()
  const expected = {
    cmd: 'publish',
    retain: false,
    qos: 0,
    dup: false,
    length: length,
    topic: topic,
    payload: Buffer.from('a'.repeat(payloadLength))
  }

  parser.on('packet', packet => {
    t.deepLooseEqual(packet, expected, 'expected packet')
  })

  t.equal(parser.parse(Buffer.from([
    48, 255, 1, // Header
    0, topic.length, // Topic length
    116, 101, 115, 116 // Topic (test)
  ])), 6, 'remaining bytes')

  t.equal(parser.parse(Buffer.from(Array(payloadLength).fill(97))),
          0, 'remaining bytes')
})

test('split length parse', t => {
  t.plan(4)

  const length = 255
  const topic = 'test'
[...]

PR #96, opened 2020-11-06
Fix: error handling for Variable Byte Integer #95

twegener-embertec:
This fixes the regression outlined in #95 which was introduced by #90 .
See #95 for more specific details of the primary problem being addressed here.

Changes:

- Avoid emitting false errors when not all bytes are present yet when parsing the variable byte integer for the length header.
- Emit error upon attempting to generate a variable byte integer beyond the maximum allowed size.
- Add more test coverage in this area.

### DIFF
--- a/parser.js
+++ b/parser.js
@@ -72,8 +72,6 @@ class Parser extends EventEmitter {
     if (result) {
       this.packet.length = result.value
       this._list.consume(result.bytes)
-    } else {
-      this._emitError(new Error('Invalid length'))
     }
     debug('_parseLength %d', result.value)
     return !!result
@@ -542,6 +540,7 @@ class Parser extends EventEmitter {
 
   _parseVarByteNum (fullInfoFlag) {
     debug('_parseVarByteNum')
+    const maxBytes = 4
     let bytes = 0
     let mul = 1
     let value = 0
@@ -549,7 +548,7 @@ class Parser extends EventEmitter {
     let current
     const padding = this._pos ? this._pos : 0
 
-    while (bytes < 5) {
+    while (bytes < maxBytes) {
       current = this._list.readUInt8(padding + bytes++)
       value += mul * (current & constants.VARBYTEINT_MASK)
       mul *= 0x80
@@ -561,6 +560,10 @@ class Parser extends EventEmitter {
       if (this._list.length <= bytes) {
         break
       }
+    }
+
+    if (!result && bytes === maxBytes && this._list.length >= bytes) {
+      this._emitError(new Error('Invalid variable byte integer'))
     }
 
     if (padding) {
--- a/constants.js
+++ b/constants.js
@@ -39,6 +39,7 @@ protocol.RETAIN_MASK = 0x01
 /* Length */
 protocol.VARBYTEINT_MASK = 0x7F
 protocol.VARBYTEINT_FIN_MASK = 0x80
+protocol.VARBYTEINT_MAX = 268435455
 
 /* Connack */
 protocol.SESSIONPRESENT_MASK = 0x01
--- a/test.js
+++ b/test.js
@@ -1283,6 +1283,16 @@ test('splitted publish parse', t => {
   ])), 0, 'remaining bytes')
 })
 
+testGenerateError('Invalid length: 268435456', {
+  cmd: 'publish',
+  retain: false,
+  qos: 0,
+  dup: false,
+  length: (268435455 + 1),
+  topic: 'test',
+  payload: Buffer.alloc(268435455 + 1 - 6)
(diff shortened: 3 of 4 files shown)


mcollina:
@YoDaMa could you take a look as well?

mcollina (approved):
lgtm

parser.js
```diff
@@ -542,14 +540,15 @@ class Parser extends EventEmitter {
...
+    const maxBytes = 4
     let bytes = 0
     let mul = 1
     let value = 0
     let result = false
     let current
     const padding = this._pos ? this._pos : 0
 
-    while (bytes < 5) {
+    while (bytes < maxBytes) {
```
YoDaMa, line 551: shouldn't this be a less than or equal for maxBytes

twegener-embertec: I believe that the previous ' bytes < 5' was incorrect.

The maximum size of a var byte int is 4 bytes, so 4 bytes need to be examined in the loop.
The bytes variable starts at 0, so to examine 4 bytes, the loop has to be over 0, 1, 2, 3.
If it doesn't meet the ((current & constants.VARBYTEINT_FIN_MASK) === 0) condition for any of those four bytes, then that indicates that it has exceeded the maximum size, i.e. that bit being non-zero means that it is expecting more length bytes, but we are already at the maximum at for bytes === 3.

Here are a couple of tests that demonstrate that buggy aspect present in v6.6.0, failing to emit the required error there:
```
testParseError('Invalid length', Buffer.from(
  [16, 255, 255, 255, 255, 1]
), {})
testParseError('Invalid length', Buffer.from(
  [16, 255, 255, 255, 255, 127]
), {})
```

While the following tests do pass in v6.6.0:
```
testParseError('Invalid length', Buffer.from(
  [16, 255, 255, 255, 255, 128]
), {})
testParseError('Invalid length', Buffer.from(
  [16, 255, 255, 255, 255, 255, 1]
), {})
```

(Note that the error message is different between v6.6.0 and my changes, i.e. 'Invalid length' vs 'Invalid variable byte integer'.)

With the updated error message, the following tests all pass with my changes:

```
testParseError('Invalid variable byte integer', Buffer.from(
  [16, 255, 255, 255, 255]
), {})
testParseError('Invalid variable byte integer', Buffer.from(
  [16, 255, 255, 255, 128]
), {})
testParseError('Invalid variable byte integer', Buffer.from(
  [16, 255, 255, 255, 255, 1]
), {})
testParseError('Invalid variable byte integer', Buffer.from(
  [16, 255, 255, 255, 255, 127]
), {})
testParseError('Invalid variable byte integer', Buffer.from(
  [16, 255, 255, 255, 255, 128]
), {})
testParseError('Invalid variable byte integer', Buffer.from(
  [16, 255, 255, 255, 255, 255, 1]
), {})
```

That seems to be good evidence that my version has the correct while condition and that the previous one was another bug in v6.6.0.

YoDaMa:
can you squash your commits down into 1 or 2 commits then I'll approve and merge @twegener-embertec

twegener-embertec:
@YoDaMa can you please explain to me your rationale / guidelines / philosophy for squashing the commits?

I deliberately kept those commits separate since the purpose/effects of each are distinct:
- Add additional test coverage that still passes against the old code.
- Add test that exposes one pre-existing bug.
- Add test that exposes another pre-existing bug.
- Provide the fix.
- Restructure the test to deal with some tangentially related nodejs compatibility issue.

I'm reluctant to squash them as that loses that information, and also makes it harder to rewind and demonstrate the historical bugginess aspects.

I don't currently see any benefits in squashing them.

Perhaps you could do a --no-ff merge if you want the overall change to be obvious as well?

YoDaMa:
> @YoDaMa can you please explain to me your rationale / guidelines / philosophy for squashing the commits?
> 
> I deliberately kept those commits separate since the purpose/effects of each are distinct:

fair. I like that logic. The last commit seems extraneous beyond the earlier 5 commits. But it's good enough.

twegener-embertec:
> The last commit seems extraneous beyond the earlier 5 commits. But it's good enough.

Assuming you mean 909cbd3626, that was added to give evidence in response to your concerns about the maxBytes threshold.
Ideally, I suppose that that commit should be reordered prior to the fix commit, so that it demonstrates that other bug before the fix.

If you want I could do that reordering via an interactive rebase, but I'd want to do that on a new branch to avoid screwing with this already public branch. I'm not sure how to integrate such a revised branch into this pull request, so if you want that please advise how. Otherwise I'll assume the "it's good enough" still stands.

psorowka:
@mcollina is this already released and included in a current mqttjs release?

PR #98, opened 2021-02-03
Fix mqtt5 disconnect can have no property fields

ogis-yamazaki:
@mcollina
@linwe2012 

I need this PR(#78), so I fixed the conflict

fix #78 conflict

thanks

### DIFF
--- a/parser.js
+++ b/parser.js
@@ -441,10 +441,12 @@ class Parser extends EventEmitter {
     if (this.settings.protocolVersion === 5) {
       // response code
       packet.reasonCode = this._parseByte()
-      // properies mqtt 5
-      const properties = this._parseProperties()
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

bkp7:
This PR doesn't address the underlying problem of the buffer overflow. The _parseProperties function is used in other places and the issue remains in those places. My PR #100 fixes the underlying cause of the out of bounds error. I have also included the test for the particular issue identified in this within my PR.

mcollina:
@bkp7 do you mean this should not land?

bkp7:
> @bkp7 do you mean this should not land?

@mcollina yes I do.

The underlying problem is with _parseVarByteNum(). If my PR #100 is accepted, this one is not required, and I have added the test to my PR to show that.

Issue #72, opened 2020-01-19
_parseVarByteNum  parser error


blue-bird1:
Will cause `RangeError [ERR_BUFFER_OUT_OF_BOUNDS]: Attempt to write outside buffer bounds in special packets

`https://gist.github.com/blue-bird1/5f79ca8665c4fc9dffb7cd3ed3233490`

bkp7:
PR #100 has been committed and in the absence of a specified test for this, can we assume this issue is now fixed and close it?

PR #100, opened 2021-02-09
Fix Buffer Out of Bounds Error

### LINKED ISSUE #99: Buffer out of bounds when v5 connection request sent to v4 server
Caused by trying to decode a V4 connack using code expecting a V5 connack (because V5 has been set in options).

Underlying issue traced to _parseVarByteNum function in parser.js which doesn't check buffer length before reading additional UInt8.

bkp7:
Fix bug in _parseVarByteNum() function causing Buffer Out of Bounds Error.

Added Tests.

Also refactored the testParseGenerateDefaults() test function to separate out a new testParseOnly() which can be used directly.

Fixes #99, fixes #98, closes #78, fixes mqttjs/MQTT.js#1073, fixes mqttjs/MQTT.js#993, fixes mqttjs/MQTT.js#1176

Possible fix to mqttjs/MQTT.js#1228 which has this function in the call stack. Unfortunately no specific test provided in that issue to prove it's fixed.

Probable fix to #72, but gist and test not longer present to check it.

### DIFF
--- a/test.js
+++ b/test.js
@@ -97,6 +97,26 @@ function testGenerateOnly (name, object, buffer, opts) {
   })
 }
 
+function testParseOnly (name, object, buffer, opts) {
+  test(name, t => {
+    const parser = mqtt.parser(opts)
+    // const expected = object
+    // const fixture = buffer
+
+    t.plan(2 + Object.keys(object).length)
+
+    parser.on('packet', packet => {
+      t.equal(Object.keys(object).length, Object.keys(packet).length, 'key count')
+      Object.keys(object).forEach(key => {
+        t.deepEqual(packet[key], object[key], `expected packet property ${key}`)
+      })
+    })
+
+    t.equal(parser.parse(buffer), 0, 'remaining bytes')
+    t.end()
+  })
+}
+
 function testParseError (expected, fixture, opts) {
   test(expected, t => {
     t.plan(1)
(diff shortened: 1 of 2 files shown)


test.js
```diff
@@ -515,13 +516,69 @@ testParseGenerate('no clientId with 3.1.1', {
...
+  retain: false,
+  qos: 0,
+  dup: false,
+  length: 16,
+  topic: null,
+  payload: null,
+  protocolId: 'MQTT',
+  protocolVersion: 4,
+  clean: true,
+  keepalive: 0,
```
mcollina, line 528: this tests had this values omitted on purpose for the generation tests, as it tests the defaults are there. It's not equivalent at the previous one.

mcollina (changes requested):

test.js
bkp7: @mcollina, yes, sorry I looked at the test as doing both a Parse and Generate and didn't twig about the defaults (the clue was in the name!). I was after a two way test so I've created a new testParseAndGenerate to do that for my new tests.

On the existing test I have added an additional field so it takes the partial packet object, generates the mqtt bytes, and then parses those to create the fully populated packet object. I believe this now fully covers the test that was previously being done but with the added benefit of checking the generated object is exactly as expected.

mcollina (approved):
lgtm

Issue #92, opened 2020-10-01
Mqtt 5 puback is different from spec


tekjar:
```
3.4.2.2.1 Property Length
The length of the Properties in the PUBACK packet Variable Header encoded as a Variable Byte Integer. If the Remaining Length is less than 4 there is no Property Length and the value of 0 is used.
```

Spec says that if remaining len is < 4, there's no property length. I think 'there's no property length' implies that it's not part of the byte stream (beacause it checks out with mosquitto)

```js
const mqtt = require('mqtt-packet');

let puback = {
  cmd: 'puback',
  messageId: 42,
}

const opts = { protocolVersion: 5 }; // default is 4. Usually, opts is a connect packet
let buf = mqtt.generate(puback, opts)
console.log(buf.toString('hex').match(/../g).map((x) => `0x${x}`).join(', '))

```
 Output of the above code is 
```
0x40, 0x03, 0x00, 0x2a, 0x00
```

Output of ack from mosquitto with no reason code and properties is this

```
[
    0x40,
    0x2,
    0x0,
    0x9,
]
```

Is the last `0x00` necessary?

mcollina:
I actually don't know. Would you like to send a PR? Can you check if this works with Mosquitto?

tekjar:
@mcollina I can try but my java script skills are close to 0

bkp7:
Looking at the output @tekjar produced from mqtt-packet the last Byte is not the Property Length:
```js
[
  0x40, 0x03, // Fixed Header (PUBACK, Remaining Length - 3 Bytes)
  0x00, 0x2a, // Packet Identifier MSB, LSB: 42
  0x00 // PUBACK Reason Code: Success
]
```
The question remains as to whether the last Byte is necessary as both Mosquitto and HiveMQ ommit it (tested against their respective online test servers today). You need to infer that the Reason Code is Success per the comment in 3.4.2.1 "The Reason Code and Property Length **_can_** be omitted......":
```js
[
  0x40, 0x02, // Fixed Header (PUBACK, Remaining Length - 2 Bytes)
  0x00, 0x2a, // Packet Identifier MSB, LSB: 42
]
```
Reading the specification I conclude that either is acceptable, as also this would be:
```js
[
  0x40, 0x04, // Fixed Header (PUBACK, Remaining Length - 4 Bytes)
  0x00, 0x2a, // Packet Identifier MSB, LSB: 42
  0x00, // PUBACK Reason Code: Success
[...]

PR #101, opened 2021-02-09
Fix reason code missing

### LINKED ISSUE #76: Error during parsing DISCONNECT packet
It seems legit to having `DISCONNECT` packet of  `0xe0 0x00`, according to [mosquitto](https://mosquitto.org/), which means the remaining length is 0.

However parser.js:439:Parser.prototype._parseDisconnect() would not check whether the packet.length > 0. Please fix this issue.

### LINKED ISSUE #80: Reason Codes seem not to be enabled
E.g. when connecting to test.mosquitto.org, no reason code is delivered in connack packet

bkp7:
In MQTT v5.0 variable length headers allow for short messages with reason code not explicitly in the message but must be assumed to be zero. These were not appearing in packet where short variable headers were received as seen from Mosquitto and HiveMQ.

Affects: disconnect, connack,  puback, pubrec, pubrel and pubcomp (the last 4 are all dealt with in _parseConfirmation function)

Added tests for all short versions of messages and two way test for the version generated by mqtt-packet

Fixes #76 (disconnect), fixes #80 (connack).

This is related to #92 in that it is the opposite (parsing) process

NB. This is dependent on (branched from) my PR #100

### DIFF
--- a/parser.js
+++ b/parser.js
@@ -244,12 +244,17 @@ class Parser extends EventEmitter {
     debug('_parseConnack')
     const packet = this.packet
 
-    if (this._list.length < 2) return null
-
+    if (this._list.length < 1) return null
     packet.sessionPresent = !!(this._list.readUInt8(this._pos++) & constants.SESSIONPRESENT_MASK)
+
     if (this.settings.protocolVersion === 5) {
-      packet.reasonCode = this._list.readUInt8(this._pos++)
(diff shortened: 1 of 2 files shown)


mcollina:
Could you please rebase?

bkp7:
Rebased

mcollina (approved):
lgtm

PR #103, opened 2021-02-11
Improved TypeScript declarations for userProperties

### LINKED ISSUE #102: Current Typescript definitions are difficult to use to read userProperties
Using TypeScript, reading user properties requires something like this (MQTT client):
```ts
client.on('message', function(topic, message, packet) {
  const pckt: IPublishPacket = Object.assign(packet);
  if (pckt.properties && pckt.properties.userProperties) {
    const userProperties: {[index: string]: string} = Object.assign(pckt.properties.userProperties);
    console.log(userProperties['test']);
  }
});
```

It would be better if the userProperties were changed from being {object} to being {[index: string]: string}.

It would also be better if the on message callback (in MQTT) required an IPublishPacket rather than the more generic Packet.

Changing both these would allow type assignments to be removed and the TypeScript code could be changed to:
```ts
client.on('message', function(topic, message, packet) {
  if (packet.properties && packet.properties.userProperties) {
    console.log(packet.properties.userProperties['test']);
  }
});
```
Linked to mqttjs/MQTT.js#1248

bkp7:
userProperties changed from being object to being {[index: string]: string} (name-value string pairs)

fixes #102

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -19,6 +19,8 @@ export declare type PacketCmd = 'auth' |
   'unsuback' |
   'unsubscribe'
 
+export declare type UserProperties = {[index: string]: string}
+
 export interface IPacket {
   cmd: PacketCmd
   messageId?: number
@@ -32,7 +34,7 @@ export interface IAuthPacket extends IPacket {
     authenticationMethod?: string,
     authenticationData?: Buffer,
     reasonString?: string,
-    userProperties?: Object,
+    userProperties?: UserProperties,
   }
 }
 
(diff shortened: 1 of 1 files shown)


scarry1992:
UserProperty can be an array, because by mqtt5 specification each userProperty can be few times in one packet. So your declaration is wrong.

bkp7:
@scarry1992, changed to allow for the array.

I also wanted to make a PR on MQTT.js to use this but it would be dependent on an version of mqtt-packet with this change in it. How do you manage that scenario?

mcollina (approved):
lgtm

scarry1992 (approved):

PR #104, opened 2021-02-12
fix to allow zero length client identifiers in MQTT v5.0

### LINKED ISSUE #68: generate function cannot handle zero-length client identifiers

bkp7:
Same as PR #71 but includes the requested test

closes #71 fixes #68

### DIFF
--- a/test.js
+++ b/test.js
@@ -519,6 +519,37 @@ testParseGenerate('no clientId with 3.1.1', {
   0, 0 // Client ID length
 ]))
 
+testParseGenerateDefaults('no clientId with 5.0', {
+  cmd: 'connect',
+  protocolId: 'MQTT',
+  protocolVersion: 5,
+  clean: true,
+  keepalive: 60,
+  properties:
+    {
+      receiveMaximum: 20
+    },
+  clientId: ''
+}, Buffer.from(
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #106, opened 2021-02-20
Add support for mosquitto bridge versions 131 and 132

phil-mitchell:
These protocols are for bridge mode of version 3.  According to https://github.com/mqtt/mqtt.org/wiki/bridge_protocol these versions set the following properties:
- ras = true
- nl = true


mcollina (approved):
lgtm

phil-mitchell:
I'm not sure it's clear that I don't have write access, so somebody will have to merge this if it's okay.

PR #108, opened 2021-05-10
add reasonCode

ogis-yamazaki:
I am using TypeScript to implement Server Redirect for MQTT V5.0.

In Server Redirect, the value of reasonCode should be used to determine whether to forward or not. (ReasonCode 156:UseAnotherServer, 157:ServerMoved)

In the Server Redirect case, the reasonCode is needed for CONNACK and DISCONNECT, so I added it to the interface.

Similarly for other packets, since reasonCode was added in MQTT V5.0, I added it to the interface.


mcollina (approved):
lgtm

PR #112, opened 2021-05-25
fix: emit errors for invalid flags

### LINKED ISSUE #111: can't comply with MQTT spec requirements about flag validation
> The Server MUST validate that the reserved flag in the CONNECT Control Packet is set to zero and disconnect the Client if it is not zero [MQTT-3.1.2-3]

If I set that flag to 1, `mqtt-packet` still parses it just fine, but doesn't emit raw flag bits, so it's impossible for a server implementation to comply with this spec requirement using `mqtt-packet`.

```
> var parser = require('mqtt-packet').parser()
> parser.on('packet', console.log)
// the `3` in the following sets the reserved flag to 1
> parser.parse(Buffer.from('100c00044d515454040300000000', 'hex'))
Packet {
  cmd: 'connect',
  retain: false,
  qos: 0,
  dup: false,
  length: 12,
  topic: null,
  payload: null,
  protocolId: 'MQTT',
  protocolVersion: 4,
  clean: true,
  keepalive: 0,
  clientId: ''
}
```

This goes for many other packet types' flags as well...for example

> Bits 3,2,1 and 0 of the fixed header in the PUBREL Control Packet are reserved and MUST be set to 0,0,1 and 0 respectively. The Server MUST treat any other value as malformed and close the Network Connection [MQTT-3.6.1-1].

`mqtt-packet` should either attach a raw buffer to the packet, or at least some fields that represent the raw flag values.

It *might* be sufficient to emit an error when any packet's flags are invalid; I think in all cases the server is just supposed to close the connection, but I'm not 100% sure.

jedwards1211:
fix #111

Includes errors for:
* all required flags in https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901023
* connect flag bit 0 != 0
* will retain and qos != 0 when will flag is 0
* connack flag bits 7-1 != 0
* publish qos > 2

TODO:
* subscribe qos > 2
* non-zero reserved bits in subscribe payload (MQTT 3.1.1)
* non-zero reserved bits in subscribe payload (MQTT 5)
* retain handling > 2
* ...anything else?

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -21,6 +21,29 @@ protocol.types = {
   15: 'auth'
 }
 
+protocol.requiredHeaderFlags = {
+  1: 0, // 'connect'
+  2: 0, // 'connack'
+  4: 0, // 'puback'
+  5: 0, // 'pubrec'
+  6: 2, // 'pubrel'
+  7: 0, // 'pubcomp'
+  8: 2, // 'subscribe'
+  9: 0, // 'suback'
+  10: 2, // 'unsubscribe'
+  11: 0, // 'unsuback'
+  12: 0, // 'pingreq'
+  13: 0, // 'pingresp'
+  14: 0, // 'disconnect'
+  15: 0 // 'auth'
+}
+
+protocol.requiredHeaderFlagsErrors = {}
+for (const k in protocol.requiredHeaderFlags) {
+  const v = protocol.requiredHeaderFlags[k]
+  protocol.requiredHeaderFlagsErrors[k] = 'Invalid header flag bits, must be 0x' + v.toString(16) + ' for ' + protocol.types[k] + ' packet'
+}
+
 /* Mnemonic => Command code */
 protocol.codes = {}
 for (const k in protocol.types) {
--- a/parser.js
+++ b/parser.js
@@ -54,9 +54,19 @@ class Parser extends EventEmitter {
   _parseHeader () {
     // There is at least one byte in the buffer
     const zero = this._list.readUInt8(0)
-    this.packet.cmd = constants.types[zero >> constants.CMD_SHIFT]
+    const cmdIndex = zero >> constants.CMD_SHIFT
+    this.packet.cmd = constants.types[cmdIndex]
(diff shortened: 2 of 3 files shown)


mcollina (approved):
lgtm, I'll bump the major.

mcollina:
Actually I might have landed this too soon... are you still working on this? In case can you open a follow-up PR with the other checks?

jedwards1211:
Ah sorry, yeah I'll try to add the other checks today

jedwards1211:
@mcollina What's the purpose of this test?  A connack code of 1 is invalid in MQTT 5...

```

testParseOnly('Version 4 CONACK in Version 5 mode', {
  cmd: 'connack',
  retain: false,
  qos: 0,
  dup: false,
  length: 2,
  topic: null,
  payload: null,
  sessionPresent: false,
  reasonCode: 1 // a version 4 return code stored in the version 5 reasonCode because this client is in version 5
}, Buffer.from([
  32, 2, // Fixed Header (CONNACK, Remaining Length)
  0, 1 // Variable Header (Session not present, Connection Refused - unacceptable protocol version)
]), { protocolVersion: 5 }) // message is in version 4 format, but this client is in version 5 mode
```

mcollina:
yes exactly

jedwards1211:
@mcollina in that case shouldn't the connection change its generator options to `protocolVersion: 4` before generating the `connack` packet, instead of `mqtt-packet` tolerating MQTT4 return codes in MQTT5 mode?

jedwards1211:
Actually I still don't understand the point of this test, because if the client wants to connect as MQTT4, why would it be trying to parse the connack as MQTT5?

jedwards1211:
Oh sorry, now I get it.  It's about a client trying to connect as MQTT5 to a server that only supports MQTT4

jedwards1211:
Kind of odd that the MQTT5 spec doesn't allow the MQTT4 return codes since that's all that an MQTT4 server can reasonably return...

PR #113, opened 2021-05-26
Strict validate subscribe flags, qos <= 2, and all reason codes

jedwards1211:


jedwards1211:
@mcollina given the weird inconsistency between MQTT 4 and 5 connack return codes, I think it's best not to validate connack return codes at all...that was the only remaining thing so this should be good to merge

jedwards1211:
Okay this should be good to go

jedwards1211:
@mcollina okay i fixed the formatting.  Btw should there be a `yarn.lock` or `package-lock.json` in the repo?

mcollina (approved):
lgtm

mcollina:
No, they are too much trouble for what they bring.

jedwards1211:
I haven't encountered that viewpoint...You've never had that situation where a CI build suddenly fails because a newer version of a dependency gets installed?

mcollina:
> I haven't encountered that viewpoint...You've never had that situation where a CI build suddenly fails because a newer version of a dependency gets installed?

Yes I did. It's very rare as I tend to avoid those dependencies.

Issue #116, opened 2021-06-12
Generate 'puback' package can't set 'qos' properties


dim0147:
Hi, I want to create 'puback' package which has 'qos' property is **1** but it always 0:
![image](https://user-images.githubusercontent.com/44487221/121761543-fbffa200-cb62-11eb-9fea-71ff4d1e507f.png)
![image](https://user-images.githubusercontent.com/44487221/121761553-0752cd80-cb63-11eb-81df-a7297ed45954.png)

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

dim0147:
> Would you like to send a Pull Request to address this issue? Remember to add unit tests.

I have make a pull request, can you check?

scarry1992:
But puback package hasn't qos field by specification..for what you need this option?

dim0147:
Currently, I'm working with Facebook sites and they use packages that required this property to be set and I need to access this field.

scarry1992:
Look, puback always has a qos=1. Use this spec description knowledge for puback processing in your own code.

PR #117, opened 2021-06-12
Issue 116, Cannot set 'qos' property for 'puback' package

dim0147:
Hi, sorry for the file, I don't have experience in write unit test so I have write the file `mqtt-issue(#116).js` which implement the test, please check, thank you

### DIFF
--- a/mqtt-issue(#116).js
+++ b/mqtt-issue(#116).js
@@ -1,0 +1,57 @@
+const mqtt = require('mqtt-packet');
+
+function getPacket(data) {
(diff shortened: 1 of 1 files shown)


YoDaMa:
@dim0147 this is because `puback` has by definition a QoS of 1. See the spec for MQTT 3.1.1:

image

Setting the QoS and reading the QoS isn't something you do on a `PUBACK` packet...

There could be added "guardrails" to stop users from adding QoS to puback for instance... Maybe it's in the Typescript? 

Either way, this is not a bug.

PR #120, opened 2021-09-06
Add mandatory payload checks

ccarcaci:


ccarcaci:
Closing because of behavioural fix was needed.

PR #122, opened 2021-09-20
fix(connect packet types): payloadFormatIndicator

YoDaMa:
The connect packet type is shown as a boolean in examples, but the type is set as a number.


mcollina (approved):
lgtm

PR #127, opened 2022-02-04
stream.emit('error', err) -> stream.destroy(err)

### LINKED ISSUE #126: Call `stream.destroy(err)` instead of `stream.emit('error', err)`
Referencing another issue commented by @mcollina:

> Oh! I forgot we were doing that in this module. This problem was fixed long ago in Node.js. We should be calling `stream.destroy(err)` instead.
> 
> Would you like to send a PR?
> 
> _Originally posted by @mcollina in https://github.com/mqttjs/mqtt-packet/issues/124#issuecomment-1028728226_

Would be happy to open a PR for this, but had a couple of questions here:
1. Wouldn't this be a breaking change? 
2. Is this truly the desired behavior? Suppose someone tries to push a CONNECT packet with an invalid protocol version. This is an error, but do we want to destroy the entire stream? Or just fail that one packet?

vishnureddy17:
close #126

### DIFF
--- a/test.js
+++ b/test.js
@@ -180,7 +180,6 @@ function testWriteToStreamError (expected, fixture) {
     const result = mqtt.writeToStream(fixture, stream)
 
     t.false(result, 'result should be false')
-    t.end()
   })
 }
 
--- a/generate.js
+++ b/generate.js
@@ -48,6 +48,10 @@ class Accumulator extends EventEmitter {
 
     return result
   }
+
+  destroy (err) {
+    if (err) this.emit('error', err)
+  }
 }
 
 module.exports = generate
--- a/writeToStream.js
+++ b/writeToStream.js
@@ -54,7 +54,7 @@ function generate (packet, stream, opts) {
     case 'auth':
       return auth(packet, stream, opts)
     default:
-      stream.emit('error', new Error('Unknown command'))
+      stream.destroy(new Error('Unknown command'))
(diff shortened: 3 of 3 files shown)


mcollina:
CI is failing

vishnureddy17:
@mcollina Hmm, I've fixed most of the failures by adding `destroy()` to `generate.js`. I don't really understand why the remaining tests are failing, I think there's some subtle behavior the old tests were assuming that are no longer true.

Any ideas? I've allowed edits by maintainers in case it's a simple thing you're able to spot. 

Thanks! 🙂

mcollina:
Unfortunately I have no time to debug this change :(

test.js
```diff
@@ -180,7 +180,6 @@ function testWriteToStreamError (expected, fixture) {
     const result = mqtt.writeToStream(fixture, stream)
 
     t.false(result, 'result should be false')
-    t.end()
```
vishnureddy17, line 183: Basically, this test was failing because `destroy()` emits `error` asynchronously. 

Removing `t.end()` here is fine because the test will automatically end when two assertions are hit and will fail the test if more than 2 are hit (per the [tape documentation](https://github.com/substack/tape#tplann))

mcollina (approved):
lgtm

vishnureddy17:
Don't have merge permission :)

PR #128, opened 2022-02-09
fix: buffer is not defined in browser

flyfishzy:
## Background
Buffer is an API in Node.js and does not exist in browsers. When upgrading to webpack 5, webpack will no longer automatically bundles core node modules (buffer/process etc.). Which will cause the `Buffer is not defined` problems.
https://github.com/mqttjs/MQTT.js/issues/1412
https://github.com/mqttjs/MQTT.js/issues/1338

## Solutions
1. Although we can add polyfills by modifying the `webpack.config.js` file. But there are also some scenarios where the `webpack.config.js` cannot be modified (such as the Openharmony development environment I‘m using).
2. The Node.js doc recommends to explicitly import or require 'buffer':
> While the Buffer class is available within the global scope, it is still recommended to explicitly reference it via an import or require statement.
https://nodejs.org/api/buffer.html
image
After that, just insall 'buffer' can fix this problem.
```
npm install buffer
```

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -1,5 +1,6 @@
 /* Protocol - protocol constants */
 const protocol = module.exports
+const { Buffer } = require('buffer')
 
 /* Command code => mnemonic */
 protocol.types = {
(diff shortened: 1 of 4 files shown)


ralrom:
Shouldn't the mqtt-packet dependency also be update to v8.0.0?

Otherwise we still have the same problem when trying to use MQTT.js in the browser.

PR #129, opened 2022-03-02
logger: debug to pino

YoDaMa:
@mcollina would this be considered a feature change?


mcollina (changes requested):
Having a global logger is a bad idea and it will lead quickly to a feature creep. Instead, let's make it a parameter and store inside the parser. You can use https://www.npmjs.com/package/abstract-logging as a default so there is no performance overhead (no need to add pino as a dependency).

PR #134, opened 2022-08-10
fix: suback fails validation of packets with QoS 128

robertsLando:
Spotted here: https://github.com/mqttjs/MQTT.js/pull/1521

### DIFF
--- a/parser.js
+++ b/parser.js
@@ -423,8 +423,8 @@ class Parser extends EventEmitter {
           return this._emitError(new Error('Invalid suback code'))
         }
       } else {
-        if (code > 2) {
-          return this._emitError(new Error('Invalid suback QoS, must be <= 2'))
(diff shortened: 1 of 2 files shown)


robertsLando:
@mcollina If LGTY please merge and make a patch release 🙏🏼

parser.js
```diff
@@ -800,7 +800,7 @@ class Parser extends EventEmitter {
   }
 
   _emitError (err) {
-    debug('_emitError')
+    debug('_emitError', err)
```
robertsLando, line 803: This is needed because when looking for this error I had no clue what was going on even by enabling debug, I just saw an emitError log that doens't say anything about why it's failing

mcollina (approved):
lgtm

robertsLando:
Grazie Matteo! 🙏🏼

PR #145, opened 2023-11-07
fix: correctly import `EventEmitter` to prevent errors with bundlers

robertsLando:
Importing EventEmitter like `const EventEmitter  = require('events')` creates errors with bundlers like esbuild, vite and so on: `TypeError: EventEmitter2 is not a constructor`


mcollina (approved):
lgtm

MARSHALL1544 (approved):

PR #148, opened 2023-12-19
fix(conn): password can be set when username unset when mqttv5

JimMoen:
see also:
> 3.1.2.8 User Name Flag
> Position: bit 7 of the Connect Flags.
>
>
>
> If the User Name Flag is set to 0, a User Name MUST NOT be present in the Payload [MQTT-3.1.2-16]. If the User Name Flag is set to 1, a User Name MUST be present in the Payload [MQTT-3.1.2-17].
>
>
>
> 3.1.2.9 Password Flag
> Position: bit 6 of the Connect Flags.
>
>
>
> If the Password Flag is set to 0, a Password MUST NOT be present in the Payload [MQTT-3.1.2-18]. If the Password Flag is set to 1, a Password MUST be present in the Payload [MQTT-3.1.2-19].
>
>
>
> Non-normative comment
>
> This version of the protocol allows the __sending of a Password with no User Name__, where MQTT v3.1.1 did not. This reflects the common use of Password for credentials other than a password.

https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901044

### DIFF
--- a/writeToStream.js
+++ b/writeToStream.js
@@ -200,7 +200,7 @@ function connect (packet, stream, opts) {
 
   // Password
   if (password != null) {
-    if (!providedUsername) {
+    if (!providedUsername && protocolVersion !== 5) {
       stream.destroy(new Error('Username is required to use password'))
       return false
     }


writeToStream.js
```diff
@@ -200,7 +200,7 @@ function connect (packet, stream, opts) {
 
   // Password
   if (password != null) {
-    if (!providedUsername) {
+    if (!providedUsername && protocolVersion !== 5) {
```
robertsLando, line 203: Could you add a comment with a link to official docs that explain why this?

robertsLando (changes requested):

writeToStream.js
JimMoen: I'm new in javascript.
Plz wait for test cases.

writeToStream.js
```diff
@@ -200,7 +200,9 @@ function connect (packet, stream, opts) {
...
+      // `username` is not required when password is present in MQTT-v5
+      // see also: https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901044
```
robertsLando, line 205: Move this comment above the `if` for consistency

robertsLando (changes requested):

robertsLando:
@mcollina ping

PR #156, opened 2025-07-30
feat: allow bridge protocol versions in packet generation

seriousme:
This PR allows bride protocol versions (MQTT protocol+128) to be generated as discussed in https://github.com/moscajs/aedes/pull/1042#pullrequestreview-3070159033

### DIFF
--- a/writeToStream.js
+++ b/writeToStream.js
@@ -105,8 +105,9 @@ function connect (packet, stream, opts) {
     return false
   } else length += protocolId.length + 2
 
-  // Must be 3 or 4 or 5
-  if (protocolVersion !== 3 && protocolVersion !== 4 && protocolVersion !== 5) {
+  // Must be 3 or 4 or 5 (+128 for bridge functionality)
+  const maskedVersion = protocolVersion < 128 ? protocolVersion : protocolVersion - 128
+  if (maskedVersion !== 3 && maskedVersion !== 4 && maskedVersion !== 5) {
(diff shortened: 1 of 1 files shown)


robertsLando (approved):
LGTM

robertsLando:
Cool enough seems this spec is not present anywhere on MQTT specs, but can be found in other MQTT modules like: https://github.com/eclipse-mosquitto/mosquitto/blob/3cbe805e71ac41a2a20cc9b2ea6b3b619f49554a/lib/send_connect.c#L153

seriousme:
Indeed, it seems to be convention and not part of the formal spec.

seriousme:
The parser already supports it:
https://github.com/mqttjs/mqtt-packet/blob/e39fb28c10628720fb50e5eb355492684ff0caaa/parser.js#L171-L173

Looking at the parser we might want to make it symmetrical and use:
```js
protocolVersion: 4
bridgeMode: true
``` 
during generation as well.
Which seems a better solution than my current fix.

robertsLando:
@seriousme yep agree, go for that

seriousme:
I will push an update later today.

seriousme:
I did some more research and the feature was already present: 

https://github.com/mqttjs/mqtt-packet/blob/e39fb28c10628720fb50e5eb355492684ff0caaa/writeToStream.js#L225-L227

So we can close this PR.

PR #161, opened 2026-07-08
fix: preserve duplicate user properties with empty-string value

spokodev:
The MQTT 5 parser aggregates repeated User Properties sharing a name using a truthiness check on the previously stored value. An empty string is a valid zero-length UTF-8 string but is falsy, so a following occurrence of the same name takes the first-insert branch and overwrites it instead of appending — silently dropping the empty value and collapsing the array to a scalar.

Repro: parsing a PUBLISH carrying user properties `a=""` then `a="b"` yields `{ a: 'b' }` instead of `{ a: ['', 'b'] }`. `generate` emits both pairs correctly; only the parser loses data, so `generate → parse` is not a round-trip.

Use an explicit `!== undefined` presence check so every value is retained, per MQTT 5.0 §3.3.2.3.7 (a User Property name may appear more than once) and §1.5.4 (UTF-8 encoded strings may be zero length). Adds a regression test.

### DIFF
--- a/parser.js
+++ b/parser.js
@@ -761,7 +761,7 @@ class Parser extends EventEmitter {
           result[name] = Object.create(null)
         }
         const currentUserProperty = this._parseByType(constants.propertiesTypes[name])
-        if (result[name][currentUserProperty.name]) {
+        if (result[name][currentUserProperty.name] !== undefined) {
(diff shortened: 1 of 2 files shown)


robertsLando (approved):

robertsLando:
@mcollina release?

PR #162, opened 2026-07-23
fix: bound DISCONNECT/AUTH reason code by packet length, not buffer length

spokodev:
### Problem

`_parseDisconnect` and `_parseAuth` decide whether a Reason Code and Property
Length are present using `this._list.length` — the number of bytes left in the
whole running parse buffer, **which includes any following pipelined packet** —
instead of `this.packet.length`, the current packet's Remaining Length.

So a spec-valid **minimal** DISCONNECT/AUTH that is not the last packet in the
`parse()` buffer reads the *next* packet's bytes as its reason code (then a bogus
Property Length), aborting the whole parse:

```js
const parser = mqtt.parser({ protocolVersion: 5 })
parser.parse(Buffer.from([0xE0, 0x00, 0xC0, 0x00])) // DISCONNECT (len 0) + PINGREQ
// throws "Invalid disconnect reason code"  — should emit [disconnect rc=0, pingreq]

parser.parse(Buffer.from([0xE0, 0x01, 0x80, 0xC0, 0x00])) // DISCONNECT rc=0x80 + PINGREQ
// throws "Cannot parse property code type"

parser.parse(Buffer.from([0xF0, 0x00, 0xC0, 0x00])) // AUTH (len 0) + PINGREQ
// throws "Invalid auth reason code"
```

A standalone minimal packet works only because `_list.length` happens to equal 0
when it is the last thing in the buffer, which is why `generate()` round-trips
(it always emits the longer `E0 02 00 00` form) and existing tests never caught it.

Per MQTT 5.0 §3.14.2.2.1 (DISCONNECT) and §3.15.2.2.1 (AUTH): the Reason Code may
be omitted when it is `0x00` with no properties (Remaining Length 0), and *"if
the Remaining Length is less than 2, then there is no Property Length and the
value of 0 is used."*

### Fix

Gate the Reason Code on `packet.length > 0` and the properties on
`packet.length >= 2`, in both methods — mirroring `_parseConfirmation`, which
already bounds its reads with `packet.length`. `_parseByte` advances the read
position without touching `packet.length`, so it stays the packet's Remaining
Length throughout. Standalone minimal packets (where the two lengths coincide)
are unchanged.

### Test

Added a parser test feeding a minimal DISCONNECT (len 0 and len 1) and AUTH
followed by a PINGREQ, asserting both packets are emitted. Fails on `main`,
passes with the fix; the full suite is green (994 passing).

### Behaviour changes (release note)

Inside a packet the whole payload is already buffered, so a read past the
Remaining Length - or past a declared Property Length - is a Malformed Packet
(MQTT-5 §4.13), never "wait for more data". Inputs that previously parsed by
borrowing bytes from the next pipelined packet, or from the packet's own body,
now emit an `error` and stop. Consumers such as MQTT.js tear down the
connection on a parser `error`, so this is not a patch-level change.

**Knowingly reverses #101.** That PR added the implied reason code for a v5
CONNACK with Remaining Length 1, pinned by `Version 5 CONNACK test 1`. A
CONNACK always carries Connect Acknowledge Flags and a reason code (MQTT-5
§3.2.2), and defaulting the missing byte to `0` tells the client an
unauthenticated connection was **accepted** - so Remaining Length below 2 is
now rejected. The 2-byte v4-format CONNACK a v4-only server sends when it
refuses a v5 CONNECT is unchanged (`Version 4 CONACK in Version 5 mode`), as is
every CONNACK at Remaining Length 3 or more. A broker emitting the 1-byte form
connects today and will be disconnected after this.

Other inputs that used to parse and now error:

- a Property Length, property value, or `messageId` running past the packet or
  past its property block - previously a silent `null` / `-1`, or a value taken
  from the next packet;
- a truncated property-length varint - previously an empty property block with
  an implied `reasonCode: 0`;
- a v5 PUBLISH, SUBACK or UNSUBACK omitting the mandatory Property Length byte
  (MQTT-5 §3.3.2.3 / §3.9.2.1.1 / §3.11.2.1.1);
- a SUBACK carrying no reason codes - previously emitted as `granted: []`. This
  one applies to **3.1.1 as well as 5.0**: §3.9.3 requires a return code per
  subscription in both, and the guard sits outside the version branch. The
  matching UNSUBACK guard is 5.0 only, because a 3.1.1 UNSUBACK has no payload;
- a SUBSCRIBE or UNSUBSCRIBE carrying no topic filters ([MQTT-3.8.3-3],
  [MQTT-3.10.3-2]), all versions - previously emitted with an empty list.

Going the other way, the pipelined DISCONNECT/AUTH/CONNACK forms this PR is
named for now parse instead of failing with `Invalid <cmd> reason code`.

**The parser also recovers after an error for the first time.** `_resetState`
left `_pos` stale, so on `master` any mid-packet error wedged the parser
permanently - the next packet's Remaining Length was computed from the stale
offset, and the stream never resynced. Measured on `dd17095`: a malformed
CONNECT followed by a valid PUBLISH yields 2 errors, 0 packets, and 11 bytes
stuck in the buffer forever. That is fixed here.

Error messages: new failures use `Malformed <cmd>, <reason>`, and the Remaining
Length varint error joins them. Messages that existed before and are not part of
this bug class keep their text.

Because this PR rewrites several error strings, matching on `err.message` is no
longer a stable contract. Every parser error now carries `err.code`
(`'MALFORMED_PACKET'`) and `err.cmd`, declared as `ParserError` in
`types/index.d.ts` - match on those.

### DIFF
--- a/parser.js
+++ b/parser.js
@@ -540,7 +540,7 @@ class Parser extends EventEmitter {
 
     if (this.settings.protocolVersion === 5) {
       // response code
-      if (this._list.length > 0) {
+      if (packet.length > 0) {
         packet.reasonCode = this._parseByte()
         if (!constants.MQTT5_DISCONNECT_CODES[packet.reasonCode]) {
           this._emitError(new Error('Invalid disconnect reason code'))
@@ -548,10 +548,12 @@ class Parser extends EventEmitter {
       } else {
         packet.reasonCode = 0
       }
-      // properies mqtt 5
-      const properties = this._parseProperties()
-      if (Object.getOwnPropertyNames(properties).length) {
-        packet.properties = properties
+      // properies mqtt 5 (only present when the remaining length is >= 2)
+      if (packet.length >= 2) {
+        const properties = this._parseProperties()
+        if (Object.getOwnPropertyNames(properties).length) {
+          packet.properties = properties
+        }
       }
     }
 
@@ -569,7 +571,7 @@ class Parser extends EventEmitter {
     }
 
     // response code
-    if (this._list.length > 0) {
+    if (packet.length > 0) {
       packet.reasonCode = this._parseByte()
       if (!constants.MQTT5_AUTH_CODES[packet.reasonCode]) {
         return this._emitError(new Error('Invalid auth reason code'))
@@ -577,10 +579,12 @@ class Parser extends EventEmitter {
     } else {
       packet.reasonCode = 0
     }
-    // properies mqtt 5
-    const properties = this._parseProperties()
-    if (Object.getOwnPropertyNames(properties).length) {
-      packet.properties = properties
+    // properies mqtt 5 (only present when the remaining length is >= 2)
+    if (packet.length >= 2) {
+      const properties = this._parseProperties()
+      if (Object.getOwnPropertyNames(properties).length) {
+        packet.properties = properties
+      }
     }
 
     debug('_parseAuth: result: true')
--- a/parser.js
+++ b/parser.js
@@ -27,10 +27,13 @@ class Parser extends EventEmitter {
   }
 
   _resetState () {
-    debug('_resetState: resetting packet, error, _list, and _stateCounter')
+    debug('_resetState: resetting packet, error, _list, _pos, and _stateCounter')
     this.packet = new Packet()
     this.error = null
     this._list = bl()
+    // Reset _pos too: a mid-packet error leaves it non-zero, and the next
+    // packet's _parseVarByteNum would use the stale value as padding.
+    this._pos = 0
     this._stateCounter = 0
   }
 
@@ -164,7 +167,7 @@ class Parser extends EventEmitter {
     packet.protocolId = protocolId
 
     // Parse constants version number
-    if (this._pos >= this._list.length) return this._emitError(new Error('Packet too short'))
+    if (this._pos >= packet.length) return this._emitError(new Error('Packet too short'))
 
     packet.protocolVersion = this._list.readUInt8(this._pos)
 
@@ -179,7 +182,7 @@ class Parser extends EventEmitter {
 
     this._pos++
 
-    if (this._pos >= this._list.length) {
+    if (this._pos >= packet.length) {
       return this._emitError(new Error('Packet too short'))
     }
 
@@ -273,7 +276,7 @@ class Parser extends EventEmitter {
     debug('_parseConnack')
     const packet = this.packet
 
-    if (this._list.length < 1) return null
+    if (packet.length < 1) return null
     const flags = this._list.readUInt8(this._pos++)
     if (flags > 1) {
       return this._emitError(new Error('Invalid connack flags, bits 7-1 must be set to 0'))
@@ -281,13 +284,15 @@ class Parser extends EventEmitter {
     packet.sessionPresent = !!(flags & constants.SESSIONPRESENT_MASK)
 
     if (this.settings.protocolVersion === 5) {
-      if (this._list.length >= 2) {
+      // Bound by packet.length, not _list.length: with a pipelined packet
+      // following, _list.length would be >= 2 and read the next packet's byte.
+      if (packet.length >= 2) {
         packet.reasonCode = this._list.readUInt8(this._pos++)
       } else {
         packet.reasonCode = 0
       }
     } else {
-      if (this._list.length < 2) return null
+      if (packet.length < 2) return null
       packet.returnCode = this._list.readUInt8(this._pos++)
     }
 
@@ -548,9 +553,9 @@ class Parser extends EventEmitter {
       } else {
         packet.reasonCode = 0
       }
-      // properies mqtt 5 (only present when the remaining length is >= 2)
+      // Properties are only present when the remaining length is >= 2 (MQTT-5 §3.14.2.2.1)
       if (packet.length >= 2) {
-        const properties = this._parseProperties(packet.length)
+        const properties = this._parseProperties()
         if (Object.getOwnPropertyNames(properties).length) {
           packet.properties = properties
         }
@@ -579,9 +584,9 @@ class Parser extends EventEmitter {
     } else {
       packet.reasonCode = 0
     }
-    // properies mqtt 5 (only present when the remaining length is >= 2)
+    // Properties are only present when the remaining length is >= 2 (MQTT-5 §3.15.2.2.1)
     if (packet.length >= 2) {
-      const properties = this._parseProperties(packet.length)
+      const properties = this._parseProperties()
       if (Object.getOwnPropertyNames(properties).length) {
         packet.properties = properties
       }
@@ -640,6 +645,9 @@ class Parser extends EventEmitter {
 
   _parseNum () {
     if (this._list.length - this._pos < 2) return -1
+    // Once packet.length is known, stay inside this packet's payload so a
+    // pipelined packet's bytes are never read as this one's.
+    if (this.packet.length !== -1 && this._pos + 2 > this.packet.length) return -1
 
     const result = this._list.readUInt16BE(this._pos)
     this._pos += 2
@@ -649,6 +657,7 @@ class Parser extends EventEmitter {
 
   _parse4ByteNum () {
     if (this._list.length - this._pos < 4) return -1
+    if (this.packet.length !== -1 && this._pos + 4 > this.packet.length) return -1
 
     const result = this._list.readUInt32BE(this._pos)
     this._pos += 4
@@ -665,8 +674,13 @@ class Parser extends EventEmitter {
     let result = false
     let current
     const padding = this._pos ? this._pos : 0
+    // Header phase (packet.length === -1) parses the Remaining Length against
+    // the whole buffer; payload phase stays inside this packet's boundary.
+    const limit = this.packet.length !== -1
+      ? Math.min(this._list.length, this.packet.length)
+      : this._list.length
 
-    while (bytes < maxBytes && (padding + bytes) < this._list.length) {
+    while (bytes < maxBytes && (padding + bytes) < limit) {
       current = this._list.readUInt8(padding + bytes++)
       value += mul * (current & constants.VARBYTEINT_MASK)
       mul *= 0x80
@@ -704,7 +718,7 @@ class Parser extends EventEmitter {
 
   _parseByte () {
     let result
-    if (this._pos < this._list.length) {
+    if (this._pos < this._list.length && (this.packet.length === -1 || this._pos < this.packet.length)) {
       result = this._list.readUInt8(this._pos)
       this._pos++
     }
@@ -742,15 +756,14 @@ class Parser extends EventEmitter {
     }
   }
 
-  _parseProperties (boundary) {
+  _parseProperties () {
     debug('_parseProperties')
     const length = this._parseVarByteNum()
     const start = this._pos
     const end = start + length
-    // When the caller passes the packet boundary (DISCONNECT/AUTH pass their
-    // remaining length), a declared property length that runs past it would
+    // A declared property length that runs past this packet's boundary would
     // read the following pipelined packet's bytes. Treat that as malformed.
-    if (boundary !== undefined && end > boundary) {
+    if (this.packet.length !== -1 && end > this.packet.length) {
       this._emitError(new Error('Property length exceeds packet length'))
       return false
     }
--- a/parser.js
+++ b/parser.js
@@ -543,14 +543,14 @@ class Parser extends EventEmitter {
       if (packet.length > 0) {
         packet.reasonCode = this._parseByte()
         if (!constants.MQTT5_DISCONNECT_CODES[packet.reasonCode]) {
-          this._emitError(new Error('Invalid disconnect reason code'))
+          return this._emitError(new Error('Invalid disconnect reason code'))
         }
       } else {
         packet.reasonCode = 0
       }
       // properies mqtt 5 (only present when the remaining length is >= 2)
       if (packet.length >= 2) {
-        const properties = this._parseProperties()
+        const properties = this._parseProperties(packet.length)
         if (Object.getOwnPropertyNames(properties).length) {
           packet.properties = properties
         }
@@ -581,7 +581,7 @@ class Parser extends EventEmitter {
     }
     // properies mqtt 5 (only present when the remaining length is >= 2)
     if (packet.length >= 2) {
-      const properties = this._parseProperties()
+      const properties = this._parseProperties(packet.length)
       if (Object.getOwnPropertyNames(properties).length) {
         packet.properties = properties
       }
@@ -742,11 +742,18 @@ class Parser extends EventEmitter {
     }
   }
 
-  _parseProperties () {
+  _parseProperties (boundary) {
     debug('_parseProperties')
     const length = this._parseVarByteNum()
     const start = this._pos
     const end = start + length
+    // When the caller passes the packet boundary (DISCONNECT/AUTH pass their
+    // remaining length), a declared property length that runs past it would
+    // read the following pipelined packet's bytes. Treat that as malformed.
+    if (boundary !== undefined && end > boundary) {
+      this._emitError(new Error('Property length exceeds packet length'))
+      return false
+    }
     const result = {}
     while (this._pos < end) {
       const type = this._parseByte()
--- a/test.js
+++ b/test.js
@@ -845,6 +845,27 @@ testParseOnly('Version 5 DISCONNECT test 2', {
 ]), { protocolVersion: 5 }
 )
 
+test('Version 5 minimal DISCONNECT/AUTH before another packet parses both', t => {
+  // A minimal DISCONNECT/AUTH (remaining length 0 or 1) must be bounded by its
+  // own remaining length, not by the whole parse buffer, so a following packet
+  // is not mis-read as its reason code or properties.
+  const cases = [
+    { bytes: [224, 0, 192, 0], cmds: ['disconnect', 'pingreq'], rcs: [0, undefined] },
+    { bytes: [224, 1, 128, 192, 0], cmds: ['disconnect', 'pingreq'], rcs: [128, undefined] },
+    { bytes: [240, 0, 192, 0], cmds: ['auth', 'pingreq'], rcs: [0, undefined] }
+  ]
+  for (const { bytes, cmds, rcs } of cases) {
+    const parser = mqtt.parser({ protocolVersion: 5 })
+    const got = []
+    parser.on('packet', p => got.push(p))
+    parser.on('error', e => t.fail(`unexpected error: ${e.message}`))
+    parser.parse(Buffer.from(bytes))
+    t.deepEqual(got.map(p => p.cmd), cmds, 'both packets parsed')
+    t.deepEqual(got.map(p => p.reasonCode), rcs, 'reason codes')
+  }
+  t.end()
+})
+
 testParseAndGenerate('Version 5 DISCONNECT test 3', {
   cmd: 'disconnect',
   retain: false,
--- a/test.js
+++ b/test.js
@@ -845,23 +845,33 @@ testParseOnly('Version 5 DISCONNECT test 2', {
 ]), { protocolVersion: 5 }
 )
 
-test('Version 5 minimal DISCONNECT/AUTH before another packet parses both', t => {
-  // A minimal DISCONNECT/AUTH (remaining length 0 or 1) must be bounded by its
-  // own remaining length, not by the whole parse buffer, so a following packet
-  // is not mis-read as its reason code or properties.
+test('Version 5 DISCONNECT/AUTH before another packet parses both at every remaining length', t => {
+  // A DISCONNECT/AUTH must be bounded by its own remaining length, not by the
+  // whole parse buffer, so a following packet is not mis-read as its reason
+  // code or properties. Remaining length 0 and 1 are the regression; 2 and
+  // above are the path this PR newly gates with `packet.length >= 2` and
+  // already worked, so they are here to keep working.
+  const pingreq = [192, 0]
   const cases = [
-    { bytes: [224, 0, 192, 0], cmds: ['disconnect', 'pingreq'], rcs: [0, undefined] },
-    { bytes: [224, 1, 128, 192, 0], cmds: ['disconnect', 'pingreq'], rcs: [128, undefined] },
-    { bytes: [240, 0, 192, 0], cmds: ['auth', 'pingreq'], rcs: [0, undefined] }
+    { what: 'disconnect rl0', body: [224, 0], cmd: 'disconnect', rc: 0 },
+    { what: 'disconnect rl1', body: [224, 1, 128], cmd: 'disconnect', rc: 128 },
+    { what: 'disconnect rl2', body: [224, 2, 0, 0], cmd: 'disconnect', rc: 0 },
+    { what: 'disconnect rl7', body: [224, 7, 0, 5, 0x11, 0, 0, 0, 10], cmd: 'disconnect', rc: 0, props: { sessionExpiryInterval: 10 } },
+    { what: 'auth rl0', body: [240, 0], cmd: 'auth', rc: 0 },
+    { what: 'auth rl1', body: [240, 1, 24], cmd: 'auth', rc: 24 },
+    { what: 'auth rl2', body: [240, 2, 0, 0], cmd: 'auth', rc: 0 },
+    { what: 'auth rl7', body: [240, 7, 24, 5, 0x1F, 0, 2, 0x68, 0x69], cmd: 'auth', rc: 24, props: { reasonString: 'hi' } }
   ]
-  for (const { bytes, cmds, rcs } of cases) {
+  for (const { what, body, cmd, rc, props } of cases) {
     const parser = mqtt.parser({ protocolVersion: 5 })
     const got = []
     parser.on('packet', p => got.push(p))
-    parser.on('error', e => t.fail(`unexpected error: ${e.message}`))
-    parser.parse(Buffer.from(bytes))
-    t.deepEqual(got.map(p => p.cmd), cmds, 'both packets parsed')
-    t.deepEqual(got.map(p => p.reasonCode), rcs, 'reason codes')
+    parser.on('error', e => t.fail(`${what}: unexpected error: ${e.message}`))
+    const remaining = parser.parse(Buffer.from([...body, ...pingreq]))
+    t.deepEqual(got.map(p => p.cmd), [cmd, 'pingreq'], `${what}: both packets parsed`)
+    t.equal(got[0] && got[0].reasonCode, rc, `${what}: reason code`)
+    t.deepEqual(got[0] && got[0].properties, props, `${what}: properties`)
+    t.equal(remaining, 0, `${what}: remaining bytes`)
   }
   t.end()
 })
--- a/test.js
+++ b/test.js
@@ -902,6 +902,91 @@ test('Version 5 DISCONNECT/AUTH property length past remaining length errors, no
   }
 })
 
+test('Version 5 CONNACK binds reason code and properties to the packet, not the buffer', t => {
+  // _parseConnack read reasonCode/properties by _list.length, so a pipelined
+  // packet's bytes leaked in (reasonCode became the next packet's header byte).
+  // Bound by packet.length; a trailing PINGREQ must survive untouched.
+  const ping = [0xC0, 0x00]
+  const cases = [
+    { bytes: [0x20, 0x01, 0x00, ...ping], rc: 0, props: undefined, what: 'remaining length 1, no reason code byte' },
+    { bytes: [0x20, 0x03, 0x00, 0x00, 0x00, ...ping], rc: 0, props: undefined, what: 'remaining length 3, empty properties' },
+    { bytes: [0x20, 0x06, 0x00, 0x00, 0x03, 0x21, 0x00, 0x0a, ...ping], rc: 0, props: { receiveMaximum: 10 }, what: 'remaining length 6, a property' }
+  ]
+  t.plan(cases.length * 5)
+  for (const c of cases) {
+    const parser = mqtt.parser({ protocolVersion: 5 })
+    const packets = []
+    parser.on('packet', p => packets.push(p))
+    parser.on('error', e => t.fail(`${c.what}: unexpected error ${e.message}`))
+    const remaining = parser.parse(Buffer.from(c.bytes))
+    t.equal(packets.length, 2, `${c.what}: connack and the trailing pingreq`)
+    t.equal(packets[0].reasonCode, c.rc, `${c.what}: reason code`)
+    t.deepEqual(packets[0].properties, c.props, `${c.what}: properties`)
+    t.equal(packets[1] && packets[1].cmd, 'pingreq', `${c.what}: pingreq intact`)
+    t.equal(remaining, 0, `${c.what}: no bytes left over`)
+  }
+})
+
+test('Version 4 CONNECT truncated before a pipelined packet errors, not over-reads', t => {
+  // The version and connect-flags gates used _list.length, so a following
+  // packet's byte was read as protocolVersion/flags instead of erroring.
+  t.plan(2)
+  const parser = mqtt.parser({ protocolVersion: 4 })
+  let errors = 0
+  parser.on('error', e => {
+    errors++
+    t.equal(e.message, 'Packet too short', 'truncated connect error')
+  })
+  parser.on('packet', () => t.fail('no packet from a truncated connect'))
+  // CONNECT remaining length 6 stops after the protocol name, then a PINGREQ
+  parser.parse(Buffer.from([0x10, 0x06, 0x00, 0x04, 77, 81, 84, 84, 0xC0, 0x00]))
+  t.equal(errors, 1, 'exactly one error')
+})
+
+test('the parser recovers _pos after a mid-packet error', t => {
+  // _resetState did not clear _pos, so the next packet's _parseVarByteNum used
+  // the stale value as padding and computed a bogus remaining length.
+  t.plan(4)
+  const parser = mqtt.parser({ protocolVersion: 4 })
+  const packets = []
+  parser.on('packet', p => packets.push(p))
+  parser.on('error', () => {})
+  // A malformed CONNECT errors mid-payload, leaving _pos non-zero
+  parser.parse(Buffer.from([0x10, 0x04, 0x00, 0x06, 0x4d, 0x51]))
+  t.equal(packets.length, 0, 'no packet from the malformed connect')
+  // A following valid PUBLISH must parse cleanly rather than wedge
+  const remaining = parser.parse(Buffer.from([48, 10, 0, 4, 116, 101, 115, 116, 116, 101, 115, 116]))
+  t.equal(packets.length, 1, 'the following publish parses')
+  t.equal(packets[0] && packets[0].topic, 'test', 'with the correct topic')
+  t.equal(remaining, 0, 'no bytes left over')
+})
+
+test('Version 5 DISCONNECT and AUTH stop at the packet boundary when pipelined', t => {
+  // The reason code and property block must be bound by the packet's remaining
+  // length so a trailing PINGREQ is not consumed as part of this packet.
+  const ping = [0xC0, 0x00]
+  const cases = [
+    { bytes: [0xE0, 0x00, ...ping], cmd: 'disconnect', rc: 0, props: undefined, what: 'DISCONNECT remaining length 0' },
+    { bytes: [0xE0, 0x02, 0x00, 0x00, ...ping], cmd: 'disconnect', rc: 0, props: undefined, what: 'DISCONNECT remaining length 2, empty properties' },
+    { bytes: [0xE0, 0x07, 0x00, 0x05, 0x11, 0x00, 0x00, 0x00, 0x0a, ...ping], cmd: 'disconnect', rc: 0, props: { sessionExpiryInterval: 10 }, what: 'DISCONNECT remaining length 7, a property' },
+    { bytes: [0xF0, 0x01, 0x18, ...ping], cmd: 'auth', rc: 0x18, props: undefined, what: 'AUTH remaining length 1' },
+    { bytes: [0xF0, 0x06, 0x18, 0x04, 0x16, 0x00, 0x01, 0xAA, ...ping], cmd: 'auth', rc: 0x18, props: { authenticationData: Buffer.from([0xAA]) }, what: 'AUTH remaining length 6, a property' }
+  ]
+  t.plan(cases.length * 5)
+  for (const c of cases) {
+    const parser = mqtt.parser({ protocolVersion: 5 })
+    const packets = []
+    parser.on('packet', p => packets.push(p))
+    parser.on('error', e => t.fail(`${c.what}: unexpected error ${e.message}`))
+    const remaining = parser.parse(Buffer.from(c.bytes))
+    t.equal(packets.length, 2, `${c.what}: packet and the trailing pingreq`)
+    t.equal(packets[0].reasonCode, c.rc, `${c.what}: reason code`)
+    t.deepEqual(packets[0].properties, c.props, `${c.what}: properties`)
+    t.equal(packets[1] && packets[1].cmd, 'pingreq', `${c.what}: pingreq intact`)
+    t.equal(remaining, 0, `${c.what}: no bytes left over`)
+  }
+})
+
 testParseAndGenerate('Version 5 DISCONNECT test 3', {
   cmd: 'disconnect',
   retain: false,
--- a/test.js
+++ b/test.js
@@ -866,6 +866,42 @@ test('Version 5 minimal DISCONNECT/AUTH before another packet parses both', t =>
   t.end()
 })
 
+test('Version 5 DISCONNECT invalid reason code emits a single error', t => {
+  // An invalid reason code must stop parsing, not fall through into the
+  // property block and emit a second error (mirrors _parseAuth's return).
+  t.plan(2)
+  const parser = mqtt.parser({ protocolVersion: 5 })
+  let errors = 0
+  parser.on('error', e => {
+    errors++
+    t.equal(e.message, 'Invalid disconnect reason code', 'reason code error')
+  })
+  parser.on('packet', () => t.fail('no packet after error'))
+  parser.parse(Buffer.from([224, 3, 0x03, 0x01, 0x99]))
+  t.equal(errors, 1, 'exactly one error')
+})
+
+test('Version 5 DISCONNECT/AUTH property length past remaining length errors, not over-reads', t => {
+  // A declared property length must stay within the packet's remaining length,
+  // otherwise _parseProperties would read the following pipelined packet.
+  const cases = [
+    [224, 3, 0, 2, 1, 192, 0], // DISCONNECT rl3: propLen 2 > 1 available, then PINGREQ
+    [240, 3, 0, 2, 1, 192, 0] // AUTH, same shape
+  ]
+  t.plan(cases.length * 2)
+  for (const bytes of cases) {
+    const parser = mqtt.parser({ protocolVersion: 5 })
+    let errors = 0
+    parser.on('error', e => {
+      errors++
+      t.equal(e.message, 'Property length exceeds packet length', 'boundary error')
+    })
+    parser.on('packet', () => t.fail('no packet after boundary error'))
+    parser.parse(Buffer.from(bytes))
+    t.equal(errors, 1, 'exactly one error')
+  }
+})
+
 testParseAndGenerate('Version 5 DISCONNECT test 3', {
   cmd: 'disconnect',
   retain: false,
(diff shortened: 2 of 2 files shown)


parser.js
```diff
@@ -540,18 +540,20 @@ class Parser extends EventEmitter {
 
     if (this.settings.protocolVersion === 5) {
       // response code
-      if (this._list.length > 0) {
+      if (packet.length > 0) {
         packet.reasonCode = this._parseByte()
         if (!constants.MQTT5_DISCONNECT_CODES[packet.reasonCode]) {
           this._emitError(new Error('Invalid disconnect reason code'))
```
robertsLando: **[Major] · Correctness**

`_emitError` here has no `return`, so execution falls through into the properties block below and can emit a **second** `error` for the same packet. `_parseAuth` (parser.js:577) and `_parseConfirmation` (parser.js:511) both use `return this._emitError(...)`.

Verified on this branch:

```js
parser.parse(Buffer.from([224, 3, 0x03, 0x01, 0x99]))
// emits 'Invalid disconnect reason code' AND 'Unknown property'
// the identical AUTH input [240, 3, ...] emits one
```

**Why:** consumers such as MQTT.js tear down the stream on the first `error`, so the second lands on a closed/removed handler. It's also an inconsistent error shape for the same malformed-packet class.

**Fix:** mirror the `_parseAuth` return discipline. Pre-existing, but it's a one-word change in a function this PR already edits.

parser.js
```diff
@@ -540,18 +540,20 @@ class Parser extends EventEmitter {
...
         }
       } else {
         packet.reasonCode = 0
       }
-      // properies mqtt 5
-      const properties = this._parseProperties()
-      if (Object.getOwnPropertyNames(properties).length) {
-        packet.properties = properties
+      // properies mqtt 5 (only present when the remaining length is >= 2)
+      if (packet.length >= 2) {
```
robertsLando, line 557: **[Major] · Security / Correctness** — flagged independently by the correctness, security and test lenses.

This guard correctly bounds *whether* properties are present, but `_parseProperties` (parser.js:747-751) then computes `end = _pos + declaredLength` with no clamp to `packet.length`. So the PR's stated invariant — bounded by packet length, not buffer length — holds only for Remaining Length < 2. Same for `_parseAuth` at line 583.

Verified at this PR's head — note this is a silent cross-packet **data leak**, not a parse error:

```
parse([E0 04 00 05 11 AA] ++ [30 0A 00 01 74 00 "SECRET"])
-> {cmd:'disconnect', length:4, properties:{sessionExpiryInterval:0xAA300A00}}
   {cmd:'publish', payload:'SECRET'}
```

`0x30 0x0A 0x00` is the following PUBLISH's fixed header, absorbed into the DISCONNECT's int32 property. `_newPacket` still consumes only `packet.length`, so the stream resyncs and the corruption is invisible. A smaller repro emits no error at all:

```js
parser.parse(Buffer.from([224, 3, 0, 2, 1, 192, 0]))
// disconnect.properties.payloadFormatIndicator === true, decoded from the pingreq's header byte
```

It's also TCP-segmentation-dependent: delivered as two chunks the same bytes error instead of parsing, so behaviour varies with how the peer happens to write.

**Why:** this is pre-existing (the base commit behaves identically), so it isn't a regression — but it is the same over-read the PR is titled for, which makes the fix look more complete than it is.

**Fix:** clamp `end` to `packet.length` inside `_parseProperties` and error when the declared length overruns; give `_parseByte`/`_parseNum`/`_parse4ByteNum`/`_parseVarByteNum` the `end > this.packet.length` guard that `_parseString`:612 and `_parseBuffer`:632 already have. Doing it there also closes the identical over-read in CONNACK, PUBLISH and SUBACK. Reasonable to split into a follow-up — but then the PR description shouldn't claim the class is fixed.

parser.js
```diff
@@ -569,18 +571,20 @@ class Parser extends EventEmitter {
...
       }
     } else {
       packet.reasonCode = 0
     }
-    // properies mqtt 5
-    const properties = this._parseProperties()
-    if (Object.getOwnPropertyNames(properties).length) {
-      packet.properties = properties
+    // properies mqtt 5 (only present when the remaining length is >= 2)
+    if (packet.length >= 2) {
```
robertsLando, line 588: **[Nit] · Design/API** — explicitly optional.

The comment restates the code. `packet.length >= 2` encodes MQTT-5 §3.15.2.2.1 ("if Remaining Length < 2, Property Length is 0") — which is the same rule `_parseConfirmation` writes as `> 2` / `> 3`, offset by its 2-byte message ID. Citing the clause would let a future reader see the two forms are one rule rather than two conventions.

test.js
```diff
@@ -845,6 +845,27 @@ testParseOnly('Version 5 DISCONNECT test 2', {
...
 
+test('Version 5 minimal DISCONNECT/AUTH before another packet parses both', t => {
+  // A minimal DISCONNECT/AUTH (remaining length 0 or 1) must be bounded by its
+  // own remaining length, not by the whole parse buffer, so a following packet
+  // is not mis-read as its reason code or properties.
+  const cases = [
+    { bytes: [224, 0, 192, 0], cmds: ['disconnect', 'pingreq'], rcs: [0, undefined] },
+    { bytes: [224, 1, 128, 192, 0], cmds: ['disconnect', 'pingreq'], rcs: [128, undefined] },
+    { bytes: [240, 0, 192, 0], cmds: ['auth', 'pingreq'], rcs: [0, undefined] }
+  ]
```
robertsLando, line 856: **[Minor] · Tests**

The boundary matrix covers 3 of 8 cells: DISCONNECT{RL0, RL1} and AUTH{RL0}, each pipelined. Missing AUTH RL1, and DISCONNECT/AUTH at RL2 and RL>2.

**Why:** RL ≥ 2 is the branch this PR newly gates with `packet.length >= 2` — the riskiest part of the change — and it's the one cell never exercised followed by another packet. It's also the region still broken (see the `_parseProperties` comment on parser.js:552), so a case there would document the remaining gap rather than hide it.

**Fix:** extend the `cases` array with RL2 (empty properties) and RL>2 (real properties) variants for both packet types, each followed by a pingreq.

The hand-rolled loop itself is fine, for the record — `testParseOnly` assumes one packet per buffer so it can't express pipelining, and ad-hoc multi-packet tests have precedent in this file (`stops parsing after first error`).

test.js
```diff
@@ -845,6 +845,27 @@ testParseOnly('Version 5 DISCONNECT test 2', {
...
+  for (const { bytes, cmds, rcs } of cases) {
+    const parser = mqtt.parser({ protocolVersion: 5 })
+    const got = []
+    parser.on('packet', p => got.push(p))
+    parser.on('error', e => t.fail(`unexpected error: ${e.message}`))
+    parser.parse(Buffer.from(bytes))
```
robertsLando, line 862: **[Minor] · DRY**

This drops the assertion every parse helper in this file makes: `t.equal(parser.parse(fixture), 0, 'remaining bytes')`.

**Why:** without it the test passes even if the parser leaves the trailing packet unconsumed — which is close to the failure mode being fixed.

**Fix:** assert the `parse()` return value. Worth considering one `test()` per fixture too, so a failure names which case broke.

robertsLando (commented):
## Review summary

**Verdict: ship with minor changes.** The fix is correct and uses the codebase's own idiom — `_parseConfirmation` (parser.js:502,524) already bounds the same decision with `packet.length > 2` / `> 3`. Four lines, no new abstraction, thresholds match MQTT-5 §3.14.2.2.1 / §3.15.2.2.1, and it ships with a regression test. 680/680 tests pass on the branch.

Everything below is either "extend the fix to its siblings" or pre-existing bugs this PR sits next to.

**Top 3 risks**
1. `_parseConnack` has the *identical* bug and is left unfixed — the PR reads as "this bug class is handled" when it isn't.
2. `_parseProperties` is still bounded by the wire-declared length, not `packet.length`, so DISCONNECT/AUTH with Remaining Length ≥ 2 still read the following packet's bytes — silently, no error.
3. `_parseDisconnect`'s `_emitError` has no `return`, so one malformed packet emits two `error` events.

### Theme — one root cause behind almost every finding

The parser has two length concepts: `_list.length` (bytes left in the whole running buffer, spanning pipelined packets) and `packet.length` (this packet's Remaining Length). `_parseString`:612 and `_parseBuffer`:632 already bound by **both**. `_parseByte`, `_parseNum`, `_parse4ByteNum`, `_parseVarByteNum` and `_parseProperties` bound by `_list.length` only — so every caller has to re-open-code its own guard (`> 2`, `> 3`, and now `> 0`, `>= 2` twice). This PR patches two call sites correctly; the primitive is still wrong, which is why the siblings below are still broken. Pushing the `packet.length` clamp down into those primitives would close the whole class at once.

---

## Findings outside the diff

**Major — parser.js:284-291,297 — `_parseConnack` has the same bug, unfixed.** Uses the same `this._list.length` conflation, then calls `_parseProperties()` unconditionally. Verified at this PR's head:

```js
// v5 CONNACK, remaining length 1, followed by PINGREQ
parser.parse(Buffer.from([0x20, 0x01, 0x00, 0xC0, 0x00]))
// -> connack.reasonCode === 192  — that's the PINGREQ's header byte

parser.parse(Buffer.from([0x20, 0x02, 0x00, 0x00, 0xC0, 0x00]))
// -> error 'Cannot parse property code type', stream dies, 4 bytes stranded
```

v4 `returnCode` is affected the same way. Consistency argues for folding it in, since the title describes the bug class rather than one function — otherwise worth a line in the description saying CONNACK is out of scope.

**Major — parser.js:29-35 — `_resetState()` does not reset `_pos`.** Only `_parsePayload`:97 and `_newPacket`:805 do. After any mid-packet error the stale `_pos` is used as `padding` in the *next* packet's `_parseVarByteNum`. Verified: after an error leaving `_pos === 3`, a following valid PUBLISH `[48,10,0,4,'test','test']` computes `packet.length === 116` — silently swallowed, then wedged waiting for bytes that never arrive. Pre-existing, and it is the answer to "can the parser recover for the next `parse()` call": no.

**Minor — parser.js:167,182 — `_parseConnect`, same class.** Gates on `this._pos >= this._list.length`, so a truncated CONNECT followed by pipelined data reads the next packet's bytes as protocolVersion/flags instead of erroring `Packet too short`.

---

## Back-compat — clean

Single-packet buffers are byte-identical before/after (verified DISCONNECT/AUTH at RL 0/1/2). Previously-erroring inputs now parse; nothing that parsed before now errors; `reasonCode` presence is unchanged. Semver-patch is defensible — the old output was bytes stolen from the next packet, not a shape anyone could depend on.

## FYI

- No maximum packet size is enforced anywhere in `parser.js`; a declared Remaining Length of `VARBYTEINT_MAX` buffers up to 256 MiB before `_parsePayload` runs. Pre-existing, not DISCONNECT/AUTH-specific.
- Performance impact is nil: both operands are O(1) property reads (`bl`'s `.length` is incrementally maintained, not computed by walking `_bufs`), and `benchmarks/parse.js` only exercises PUBLISH.

## Coverage

Specialists run: correctness, security, DRY/codebase-fit, design/API/back-compat, tests, performance. Clean on: performance. Not run: operability, readability (no files in their lanes). Reviewed against a worktree at 3dd49ba; all reproductions above were executed there, not inferred.

spokodev:
Thanks for the deep review — genuinely useful. Pushed the two in-scope changes:

- `_parseDisconnect` now `return`s on the invalid reason code, so it no longer falls through into the property block and emits a second error. Matches `_parseAuth`.
- `_parseProperties` takes an optional boundary; `_parseDisconnect`/`_parseAuth` pass their remaining length, so a declared property length that overruns the packet now errors (`Property length exceeds packet length`) instead of reading the next pipelined packet. That closes the RL ≥ 2 over-read for both DISCONNECT and AUTH.

Full suite green (686/686, `standard` clean). Two new tests cover the single-error and the boundary cases. One existing test used an over-long DISCONNECT property length to reach `Cannot parse property code type`; it now hits the boundary guard first, so I nudged its input to stay within remaining length and keep exercising that path.

On the rest: you're right that the clean fix is pushing the `packet.length` clamp down into the primitives (`_parseByte`/`_parseNum`/`_parseProperties`) rather than re-guarding at each call site — that would also cover `_parseConnack`, `_parseConnect`, and the `_resetState` `_pos` recovery you found. I kept this PR to DISCONNECT/AUTH to match its title. Happy to open a separate follow-up PR for the primitive-level clamp + the CONNACK/CONNECT/`_pos` items (with your repros as tests) if you'd like — just say the word.

robertsLando:
@spokodev please do it here and not in a follow up

spokodev:
Done in this PR. Pushed the `packet.length` clamp down into the primitives (`_parseByte`/`_parseNum`/`_parse4ByteNum`/`_parseVarByteNum`/`_parseProperties`), payload-phase only — header phase is detected by `packet.length === -1`, so parsing the Remaining Length itself is untouched. `_parseProperties` now self-clamps and the DISCONNECT/AUTH `boundary` argument is gone.

Plus the three siblings, each with a pipelining regression test:

- **`_parseConnack`** — bound reasonCode/properties by `packet.length`. `[0x20,0x01,0x00,0xC0,0x00]` now yields `connack rc=0` + PINGREQ instead of `rc=192`.
- **`_parseConnect`** — version/flags gates use `packet.length`; a truncated CONNECT before a pipelined packet errors `Packet too short`.
- **`_resetState`** — resets `_pos`, so a mid-packet error no longer strands it into the next packet's `_parseVarByteNum`.

Boundary matrix extended to DISCONNECT/AUTH at RL 0/1/2/>2, each followed by a PINGREQ that must survive with `parse() === 0`. Single-packet buffers stay byte-identical; 732/732 green, `standard` clean. Added the MQTT-5 §3.14.2.2.1 / §3.15.2.2.1 citations too.

spokodev:
Pushed the two remaining test items in 9ab9880.

**Boundary matrix — 3 cells to 8.** DISCONNECT and AUTH at remaining length 0, 1, 2 and above, each followed by a PINGREQ. Against upstream master, `rl0` and `rl1` fail for both packet types (`Invalid disconnect reason code`, `Cannot parse property code type`) while `rl2` and `rl7` pass — so the first four cells are the regression and the last four are guards on the `packet.length >= 2` branch this PR newly gates, which is the part you flagged as riskiest.

**`parse()` return value** is now asserted per case, so a cell that leaves the trailing packet unconsumed fails instead of passing quietly.

758/758 green, `standard` clean.

The two Major items from your review went in on the 30 Jul push, in case those threads scrolled past — `_parseProperties` self-clamps at parser.js:766 (`Property length exceeds packet length`), and the primitives are bounded at 650/660/679. The MQTT-5 §3.14.2.2.1 / §3.15.2.2.1 citations are at parser.js:556 and 587.

parser.js
spokodev: Fixed — the check returns now, in `_parseDisconnect` and `_parseAuth` both.

Measured on `master` (`dd17095`) with `[224, 3, 0x03, 0x01, 0x99]`: **two** errors, `Invalid disconnect reason code` followed by `Unknown property`, because execution fell through into the property block exactly as you said. On this head: one error, no packet. Covered by `Version 5 DISCONNECT invalid reason code emits a single error`.

parser.js
spokodev: Fixed. Your PoC, measured on both sides.

`master` (`dd17095`):
```
input  : e004000511aa300a00017400534543524554
packets: disconnect, publish
errors : (none)
disconnect.properties.sessionExpiryInterval = 0xAA300A00
remaining: 0
```
this head:
```
packets: (none)
errors : Property length exceeds packet length
remaining: 16   (of 18 bytes in)
```
Payload reads are bounded by `packet.length` rather than the running buffer, so a property length that runs past the packet raises instead of swallowing the next packet's header.

Two things worth flagging rather than leaving for you to find:

1. The regression test uses the same shape but not your exact bytes — `[224, 3, 0, 2, 1, 192, 0]` (DISCONNECT rl3, declared property length 2 with 1 byte available, then PINGREQ) and the AUTH twin, in `Version 5 DISCONNECT/AUTH property length past remaining length errors, not over-reads`. Say the word if you would rather see your buffer in the file verbatim.
2. Behaviour change: the malformed DISCONNECT is now an error and parsing stops, so the stream does **not** resync past it the way you described — 16 bytes are left unconsumed. That seemed right for a protocol violation, but it is a change, so it is your call.

758/758 tests pass.

parser.js
spokodev: Done — it cites the clause instead of restating the code:

```js
// Properties are only present when the remaining length is >= 2 (MQTT-5 §3.15.2.2.1)
```
and the DISCONNECT side points at §3.14.2.2.1, so the two read as one rule rather than two conventions.

test.js
spokodev: Extended to all 8 cells: DISCONNECT and AUTH at rl0, rl1, rl2 and rl7, each followed by a pingreq — `Version 5 DISCONNECT/AUTH before another packet parses both at every remaining length`. rl7 carries real properties (`sessionExpiryInterval` on the DISCONNECT, `reasonString` on the AUTH) so the rl>2 path is exercised with content rather than an empty property block.

The region you called out as still broken is now its own test rather than a documented gap — see the reply on parser.js.

test.js
spokodev: Restored — each of the 8 cases asserts `t.equal(remaining, 0, ...)` alongside the packet list and the reason code, the way the other parse helpers in this file do.

parser.js
```diff
@@ -661,8 +674,13 @@ class Parser extends EventEmitter {
     let result = false
     let current
     const padding = this._pos ? this._pos : 0
+    // Header phase (packet.length === -1) parses the Remaining Length against
+    // the whole buffer; payload phase stays inside this packet's boundary.
+    const limit = this.packet.length !== -1
```
robertsLando: **[Blocker] · Correctness / Security**

When a property-length varint is truncated by the packet boundary, the new `limit` makes the loop exit before `bytes === maxBytes`, so the existing `Invalid variable byte integer` guard never fires and `_parseVarByteNum` returns `false`. `_parseProperties` then computes `end = start + false === start`, skips the loop and returns `{}` — no error.

Verified against `dd17095`:
```
E0 02 00 80 | C0 00     master: error "Cannot parse property code type"
                        PR:     disconnect{reasonCode:0} + pingreq, 0 errors
F0 02 18 80 | C0 00     same (AUTH)
20 03 00 00 80 | C0 00  same (CONNACK)
```

**Why:** MQTT-5 §1.5.5 requires Malformed Packet here. The PR trades an over-read for silent acceptance of a malformed packet — strictly worse for the caller, who now acts on a fabricated `reasonCode: 0` with nothing to log.

**Fix:** Distinguish "ran out of bytes inside this packet" (malformed → error) from "need more bytes from the network" (wait for more data).

parser.js
```diff
@@ -273,21 +276,23 @@ class Parser extends EventEmitter {
...
     if (flags > 1) {
       return this._emitError(new Error('Invalid connack flags, bits 7-1 must be set to 0'))
     }
     packet.sessionPresent = !!(flags & constants.SESSIONPRESENT_MASK)
 
     if (this.settings.protocolVersion === 5) {
-      if (this._list.length >= 2) {
+      // Bound by packet.length, not _list.length: with a pipelined packet
+      // following, _list.length would be >= 2 and read the next packet's byte.
+      if (packet.length >= 2) {
```
robertsLando: **[Blocker] · Correctness / Security**

A truncated v5 CONNACK now reads as Success. The `packet.length >= 2` gate defaults the reason code rather than rejecting, and the `return null` paths at :279 and :295 emit nothing at all.

Verified against `dd17095`:
```
20 00 | C0 00      master: error "Invalid connack flags"
                   PR:     connack{} — no sessionPresent, no reasonCode, 0 errors
20 01 00 | C0 00   master: connack{reasonCode:192}  (garbage from next packet)
                   PR:     connack{reasonCode:0}    (= Success)
```
Also: the `returnCode === -1 || reasonCode === -1` check at :299 is dead — `readUInt8` never returns -1.

**Why:** The CONNACK reason code *is* the authentication result. A truncated CONNACK reading as "connection accepted" is an authn gap, not a parse quirk — and remaining length 0 regressed from a hard error to silence. MQTT-5 §3.2.2 requires a 3-byte variable header.

**Fix:** Reject v5 CONNACK with `packet.length < 3` instead of defaulting the reason code; make both `return null` sites emit an error.

test.js
```diff
@@ -845,6 +845,158 @@ testParseOnly('Version 5 DISCONNECT test 2', {
...
+  }
+})
+
+test('Version 5 CONNACK binds reason code and properties to the packet, not the buffer', t => {
+  // _parseConnack read reasonCode/properties by _list.length, so a pipelined
+  // packet's bytes leaked in (reasonCode became the next packet's header byte).
+  // Bound by packet.length; a trailing PINGREQ must survive untouched.
+  const ping = [0xC0, 0x00]
+  const cases = [
+    { bytes: [0x20, 0x01, 0x00, ...ping], rc: 0, props: undefined, what: 'remaining length 1, no reason code byte' },
```
robertsLando: **[Blocker] · Tests**

This case asserts `rc: 0` for a CONNACK with remaining length 1 — locking in the behavior flagged at parser.js:289. A 1-byte CONNACK has no reason code byte; reporting Success is not the spec-correct outcome to pin.

**Why:** A test that encodes the wrong expectation will block the fix and make the regression look intentional to the next reader.

**Fix:** Expect a malformed-packet error for v5 CONNACK below remaining length 3. (Remaining length 0 isn't covered at all — worth adding.)

parser.js
```diff
@@ -636,6 +645,9 @@ class Parser extends EventEmitter {
 
   _parseNum () {
     if (this._list.length - this._pos < 2) return -1
+    // Once packet.length is known, stay inside this packet's payload so a
+    // pipelined packet's bytes are never read as this one's.
+    if (this.packet.length !== -1 && this._pos + 2 > this.packet.length) return -1
```
robertsLando: **[Major] · Correctness / Security**

`_parseNum` returns `-1` here, but `_parseMessageId` (:604) tests `packet.messageId === null`. The new bound makes `-1` the normal truncation outcome, so truncated PUBACK/PUBREC/PUBREL/PUBCOMP/SUBACK/UNSUBACK/PUBLISH-qos>0 are emitted as valid packets with no error.

`40 01 00 | C0 00` → `puback{messageId:-1}` + `pingreq`, 0 errors (master gave `messageId: 192`).

**Why:** Better than master's cross-packet bleed, but the parser still hands a forged, spec-illegal ack upstream instead of rejecting it — MQTT.js keys its outgoing store on `messageId`.

**Fix:** Make the sentinel checkable: treat `-1` as failure at `_parseMessageId` and the other `=== null` guards on `_parseNum` results.

parser.js
```diff
@@ -645,6 +657,7 @@ class Parser extends EventEmitter {
 
   _parse4ByteNum () {
     if (this._list.length - this._pos < 4) return -1
+    if (this.packet.length !== -1 && this._pos + 4 > this.packet.length) return -1
```
robertsLando: **[Major] · Tests**

The `_parseNum` / `_parse4ByteNum` half of this fix ships with no regression test, although it covers `messageId`, CONNECT `keepalive`, and every 2-/4-byte MQTT 5 property across all packet types — not just DISCONNECT/AUTH.

**Why:** Dropping just these two guards in a bad merge would pass the entire new suite while silently corrupting messageIds and keepalive on pipelined malformed packets.

**Fix:** Add a pipelined-boundary test for a truncated PUBACK/SUBACK, and one for a 4-byte v5 property value straddling the packet boundary.

parser.js
```diff
@@ -636,6 +645,9 @@ class Parser extends EventEmitter {
 
   _parseNum () {
     if (this._list.length - this._pos < 2) return -1
+    // Once packet.length is known, stay inside this packet's payload so a
+    // pipelined packet's bytes are never read as this one's.
```
robertsLando: **[Major] · DRY**

This bound is now spelled two ways across seven readers: `this.packet.length !== -1 && <pos+n> > this.packet.length` in five new places (:650, :660, :679-681, :721, :766), versus the pre-existing `end > this.packet.length` with no sentinel in `_parseString` (:617) and `_parseBuffer` (:637).

The `!== -1` half is also dead in `_parseNum`, `_parse4ByteNum` and `_parseByte` — they only run from `_parsePayload`, after `packet.length` is set — so the sentinel reads as load-bearing where it isn't.

**Why:** Two spellings of "am I still inside this packet" is precisely how the original bug survived; the next boundary fix has to find all seven call sites.

**Fix:** One private guard that owns the sentinel (e.g. `_hasBytes(n)` / `_packetEnd()`), used by every byte reader including the two pre-existing ones.

parser.js
```diff
@@ -743,6 +761,12 @@ class Parser extends EventEmitter {
     const length = this._parseVarByteNum()
     const start = this._pos
     const end = start + length
+    // A declared property length that runs past this packet's boundary would
+    // read the following pipelined packet's bytes. Treat that as malformed.
+    if (this.packet.length !== -1 && end > this.packet.length) {
```
robertsLando, line 790: **[Major] · Operability**

One boundary violation gets a real error here, while its siblings stay silent: past-boundary `_parseString`/`_parseBuffer` return `null` and `_parseNum`/`_parse4ByteNum` return `-1`, and the packet is still emitted.

Verified (unchanged from master):
```
E0 05 00 03 16 00 05 | C0 00     → disconnect{authenticationData: null}, no error
E0 06 00 04 26 00 01 61 | C0 00  → disconnect{userProperties:{a:null}}, no error
```

**Why:** Downstream cannot distinguish a truncated value from an absent one, and the single new error string makes the contract look enforced when it holds in one place out of four.

**Fix:** Pick one boundary policy: every past-boundary read errors, or none does and truncation is documented.

parser.js
```diff
@@ -743,6 +761,12 @@ class Parser extends EventEmitter {
...
+      this._emitError(new Error('Property length exceeds packet length'))
```
robertsLando: **[Minor] · Operability**

This message omits the packet cmd that :514 carries (`'Invalid ' + this.packet.cmd + ' reason code'`), and parser errors carry no `code` field.

**Why:** Callers are left matching on message strings, and can't tell which packet type failed.

**Fix:** Name the cmd in the message, as the sibling error does.

parser.js
```diff
@@ -743,6 +761,12 @@ class Parser extends EventEmitter {
...
+      return false
```
robertsLando, line 792: **[Major] · Correctness / Security**

`_parseProperties` returns `false` on this new error path, but no caller checks the return: all ~9 call sites run `Object.getOwnPropertyNames(false)` and keep parsing (`_parsePublish` still slices a payload).

**Why:** It only works today because `_emitError` sets `this.error` and `parse()`'s loop short-circuits — the guard's safety rests on an unrelated invariant in a different function.

**Fix:** Short-circuit the callers on a falsy return, or make `_parseProperties` signal failure the way its callers already handle.

robertsLando (commented):
## Code review — bound DISCONNECT/AUTH by packet length

**Verdict:** Needs work — the approach is right, but two regressions against `master` (`dd17095`) block it.

The underlying bug is real and well-diagnosed, the fix direction is correct, and the new tests are genuine regression tests rather than tautologies — 33+ assertions verifiably fail against unfixed code. All five earlier review threads were addressed, with spec citations added. Full suite passes 758/758.

### Theme (root cause of most findings)

The PR converts *over-read into the next packet* into *stop at the packet boundary* — but each stop returns `false` / `-1` / `null` into a caller that doesn't test for it. **Over-read became silent-accept.** One fix at the signalling layer closes most of the list below.

### Top 3 risks

1. A malformed property-length varint that used to error is now silently accepted as a clean packet (parser.js:679).
2. A truncated v5 CONNACK now reports `reasonCode: 0` (Success), and a new test locks that in (parser.js:289, test.js:921).
3. Every new boundary stop returns a sentinel no caller checks (parser.js:650, :768).

### Findings not anchorable to a diff line

- **Minor · correctness** — `parser.js:701`: `if (padding) this._pos += bytes` skips the advance when `_pos === 0`, so after a failed messageId the property-length byte is re-read as payload. Now reachable because of the new bound: `B0 01 00 | C0 00` → `unsuback{granted:[0]}` on this PR vs `granted:[]` on master — the PR fabricates a granted code that master didn't.
- **Minor · operability** — `parser.js:774`: a truncated int16 property surfaces as `Cannot parse property code type` (`E0 04 00 02 21 00 | C0 00`) because `_parseNum` returns `-1` without consuming and the loop re-reads. Points the reader at the identifier byte, not the short value.
- **Minor · operability** — the new `Property length exceeds packet length` tests assert one error and no packet, but nothing asserts the parser is still usable afterwards. It does recover; it's untested. The `_pos` reset fix has such a test.
- **FYI** — `parser.js:302`: v5 `_parseConnack` still calls `_parseProperties` unconditionally; spec says remaining length < 3 has no Property Length. Inconsistent with the gates added to DISCONNECT/AUTH.
- **FYI** — `_parseConfirmation` (:504-536) is a third copy of the reason-code-plus-properties block that `_parseDisconnect` and `_parseAuth` now duplicate; the offset (0 vs 2) is the only real variance.

### FYI — out of scope, worth knowing

- **Blast radius:** packets that previously parsed (by over-reading the next pipelined packet) now emit errors, and MQTT.js tears down the connection on a parser `error`. The repo has no CHANGELOG and the README documents no error contract — worth a release note.
- **Pre-existing, untouched:** `parser.js:96` — `_parsePayload` waits for `_list.length >= packet.length` with no cap, so a header declaring `FF FF FF 7F` pins up to 256 MB of heap per connection. A `maxRemainingLength` option would be the natural home.
- **Pre-existing:** after an error, `parse()` returns a non-zero remaining count but `_resetState` discards those bytes on the next call. The PR's new error paths make it reachable more often.

### Coverage

Reviewed for correctness, security, DRY/codebase fit, tests, operability and readability. Performance and API/back-compat lanes had no files. Every blocker and major claim was re-verified against pristine checkouts of `9ab9880` and `dd17095`.

parser.js
```diff
@@ -700,7 +718,7 @@ class Parser extends EventEmitter {
 
   _parseByte () {
     let result
-    if (this._pos < this._list.length) {
+    if (this._pos < this._list.length && (this.packet.length === -1 || this._pos < this.packet.length)) {
```
robertsLando: **[Minor] · Readability**

This compound guard has no rationale comment, while the same sentinel-boundary pattern added to `_parseNum` (:648-649), `_parse4ByteNum` (:659) and `_parseVarByteNum` (:677) each got one.

**Why:** A reader landing on `_parseByte` first sees a dense boolean with no explanation, inconsistent with how the rest of the PR documents this exact pattern.

**Fix:** Add the same one-line rationale, or factor the shared condition into the named helper suggested at :649 so the intent is stated once.

test.js
```diff
@@ -845,6 +845,158 @@ testParseOnly('Version 5 DISCONNECT test 2', {
 ]), { protocolVersion: 5 }
 )
 
+test('Version 5 DISCONNECT/AUTH before another packet parses both at every remaining length', t => {
+  // A DISCONNECT/AUTH must be bounded by its own remaining length, not by the
```
robertsLando: **[Minor] · Readability**

The new byte arrays carry no inline annotation and mix decimal with hex in one literal (`[224, 7, 0, 5, 0x11, 0, 0, 0, 10]`), while every pre-existing array in this file annotates each group — e.g. `224, 0 // Fixed Header (DISCONNECT, Remaining Length)`.

**Why:** A reader has to know the MQTT opcodes (224, 240, 0x20, 0xC0) and count bytes by hand to tell Remaining Length from reason code from property bytes.

**Fix:** Add the same inline `// Fixed Header (...)` / `// reason code` / `// property: ...` comments, at least on the denser cases.

test.js
```diff
@@ -845,6 +845,158 @@ testParseOnly('Version 5 DISCONNECT test 2', {
...
+    const remaining = parser.parse(Buffer.from([...body, ...pingreq]))
+    t.deepEqual(got.map(p => p.cmd), [cmd, 'pingreq'], `${what}: both packets parsed`)
+    t.equal(got[0] && got[0].reasonCode, rc, `${what}: reason code`)
+    t.deepEqual(got[0] && got[0].properties, props, `${what}: properties`)
+    t.equal(remaining, 0, `${what}: remaining bytes`)
+  }
+  t.end()
+})
+
+test('Version 5 DISCONNECT invalid reason code emits a single error', t => {
```
robertsLando: **[Minor] · DRY**

These hand-rolled error tests reimplement `testParseError` (:120), whose `t.plan(1)` already proves "exactly one error and no trailing packet" — tape fails a test whose assertion count exceeds its plan.

**Why:** The manual `errors++` counters add ~45 lines outside the file's dominant style for a guarantee the existing helper already gives.

**Fix:** Collapse to `testParseError(msg, buffer, { protocolVersion: 5 })` calls; keep a bespoke `test()` only where two packets must be observed.

test.js
```diff
@@ -845,6 +845,158 @@ testParseOnly('Version 5 DISCONNECT test 2', {
...
+  parser.parse(Buffer.from([0x10, 0x04, 0x00, 0x06, 0x4d, 0x51]))
+  t.equal(packets.length, 0, 'no packet from the malformed connect')
+  // A following valid PUBLISH must parse cleanly rather than wedge
+  const remaining = parser.parse(Buffer.from([48, 10, 0, 4, 116, 101, 115, 116, 116, 101, 115, 116]))
+  t.equal(packets.length, 1, 'the following publish parses')
+  t.equal(packets[0] && packets[0].topic, 'test', 'with the correct topic')
+  t.equal(remaining, 0, 'no bytes left over')
+})
+
+test('Version 5 DISCONNECT and AUTH stop at the packet boundary when pipelined', t => {
```
robertsLando: **[Minor] · DRY**

This table is the same test as the one at :848 — DISCONNECT/AUTH at several remaining lengths followed by a PINGREQ, asserting cmd list, reasonCode, properties and remaining.

**Why:** ~60 duplicated lines under two names whose titles don't make the distinction obvious; drift between them is inevitable.

**Fix:** Keep the :848 table (it covers strictly more remaining-length values) and fold this one's `authenticationData` case into it.

test.js
```diff
@@ -845,6 +845,158 @@ testParseOnly('Version 5 DISCONNECT test 2', {
...
+  ]
+  t.plan(cases.length * 5)
+  for (const c of cases) {
+    const parser = mqtt.parser({ protocolVersion: 5 })
+    const packets = []
+    parser.on('packet', p => packets.push(p))
+    parser.on('error', e => t.fail(`${c.what}: unexpected error ${e.message}`))
+    const remaining = parser.parse(Buffer.from(c.bytes))
+    t.equal(packets.length, 2, `${c.what}: packet and the trailing pingreq`)
+    t.equal(packets[0].reasonCode, c.rc, `${c.what}: reason code`)
```
robertsLando: **[Major] · Tests**

`packets[0]` is dereferenced without the `packets[0] &&` guard used by the sibling assertion at :934.

**Why:** When this assertion fails it throws an uncaught `TypeError` that kills the `node --test` process, so every test declared after it in the file silently never runs — confirmed by running the new suite against pre-fix `parser.js`.

**Fix:** Guard it, matching :934/:935.

test.js
```diff
@@ -3038,7 +3190,7 @@ testParseError('Malformed Subscribe Payload', Buffer.from([
 
 test('Cannot parse property code type', t => {
   const packets = Buffer.from([
-    16, 16, 0, 4, 77, 81, 84, 84, 5, 2, 0, 60, 3, 33, 0, 20, 0, 0, 98, 2, 211, 1, 224, 2, 0, 32
+    16, 16, 0, 4, 77, 81, 84, 84, 5, 2, 0, 60, 3, 33, 0, 20, 0, 0, 98, 2, 211, 1, 224, 3, 0, 1, 0
```
robertsLando, line 3628: **[Nit] · Readability**

The trailing bytes of this pre-existing test changed from `224, 2, 0, 32` to `224, 3, 0, 1, 0` with no note.

**Why:** Given this PR adds a `Property length exceeds packet length` check, it isn't obvious whether the tweak was needed to keep exercising the original `Cannot parse property code type` path or is unrelated.

**Fix:** A one-line comment on why the bytes changed saves the next reader from re-deriving it.

parser.js
robertsLando: Partly pushing back, with evidence.

Fixed, exactly as described:
- `20 00 | C0 00` -> `Packet too short`. The `return null` was emitting an empty `connack{}`; it now errors, as `master` did.
- `20 01 00 | C0 00` -> `connack{reasonCode: 0}` + `pingreq`, and the `0xC0` is never read as this packet's byte. The over-read is gone.
- `returnCode === -1 || reasonCode === -1` deleted; `readUInt8` never returns -1, as you said.
- v5 properties gated on `packet.length >= 3`, which was the FYI in your summary.

Not done: rejecting v5 CONNACK below remaining length 3. Two tests on `master` (`dd17095`) document the shorter forms on purpose:

- test.js:770 `Version 5 CONNACK test 1`, remaining length 1, with the comment *"Implied Reason code: Success, Implied no properties"*.
- test.js:690 `Version 4 CONACK in Version 5 mode`, remaining length 2 — a v4-only server refusing a v5 CONNECT answers in v4 format, and the comment says so.

So the 1- and 2-byte forms are deliberate leniency that predates this PR, not something it introduced. Rejecting them is a real behaviour change for that interop case and belongs in its own PR, not one about pipelining boundaries — so I left it alone rather than deleting two documented tests inside a fix PR.

No `reasonCode` is fabricated any more: below remaining length 2 the packet is rejected, and at 2 the byte is real. Say the word if you want the strict `< 3` rule and I will make it a follow-up.

test.js
robertsLando: Remaining length 0 is now covered and asserts an error — `testParseError('Packet too short', [0x20, 0x00, ...PINGREQ])`.

The remaining-length-1 case keeps `rc: 0`, and the reason is on the parser.js:289 thread: test.js:770 on `master` already pins that exact expectation with the comment *"Implied Reason code: Success, Implied no properties"*. The new case now carries a comment pointing at it, so the next reader sees it is the repo's existing choice rather than something this PR invented.

Leaving this open — if you want the strict `< 3` rule, this test and test.js:770 change together.

parser.js
robertsLando: Picked "every past-boundary read errors".

`_overruns(n)` is the one boundary check now, and it records the stop; `_parseProperties` turns it into an error naming the property and the packet type. Your two cases:

```
E0 05 00 03 16 00 05 | C0 00     -> Property authenticationData exceeds disconnect packet length
E0 06 00 04 26 00 01 61 | C0 00  -> Property userProperties exceeds disconnect packet length
```

Same for int16, int32, string and pair values, and for a truncated property-length varint. Reads outside the property block keep their existing, more specific messages (`Cannot parse topic`, `Cannot parse protocolId`, `Packet too short`) because those callers already check the sentinel — so the policy is uniform in what it does, not in what it says. Each has a `testParseError` case.

robertsLando:
Round 2 pushed as `0cc2666`. 765/765 pass, `standard` clean, `testRandom` error rate unchanged (90.19% vs 90.03% before — same noise floor, exit 0).

**The theme you named** — *"each stop returns `false` / `-1` / `null` into a caller that does not test for it"* — is the thing that got fixed, at the signalling layer:

- `_overruns(n)` is the single boundary check, replacing five new open-coded spellings and the two pre-existing ones in `_parseString`/`_parseBuffer`.
- Inside a packet the payload is fully buffered, so a read past the remaining length is a Malformed Packet (MQTT-5 §1.5.5), never "wait for more data". Truncated property-length varints and truncated property values now error instead of yielding `{}` or `null`.
- `_readProperties()` reads the block and returns false on failure; all eleven call sites stop on it.
- `_parseMessageId` tests `-1`.

**From the summary, not anchored to a line:**

- `parser.js:701` — `if (padding) this._pos += bytes` is now `if (this.packet.length !== -1)`. It advances from offset 0 in the payload phase, and still does not advance during the header phase, which re-runs from 0 until the whole varint arrives. Your `B0 01 00 | C0 00` case is rejected with `Cannot parse messageId` rather than fabricating `granted: [0]`.
- `parser.js:774` — `E0 04 00 02 21 00 | C0 00` now says `Property receiveMaximum exceeds disconnect packet length` instead of pointing at the identifier byte.
- The recovery gap — `the parser stays usable after a packet-boundary error` runs three malformed inputs (mid-packet CONNECT error, property length past the boundary, property value past the boundary) and asserts a valid PUBLISH parses after each.
- `parser.js:302` — v5 CONNACK properties are gated on remaining length >= 3 now.
- `_parseConfirmation` as a third copy — the properties half is shared via `_readProperties`; the reason-code half still differs by offset in all three, so I left it.

**Still open**, both about the same decision: parser.js:289 and test.js:921. `master` has two tests that deliberately accept v5 CONNACK at remaining length 1 and 2 (`Version 5 CONNACK test 1`, `Version 4 CONACK in Version 5 mode`), so the strict `< 3` rejection would delete documented behaviour inside a fix PR. Details and the exact before/after on those threads — your call.

**Out of scope, as flagged:** `maxRemainingLength`, and `parse()` returning a non-zero remaining count that `_resetState` then discards.

parser.js
robertsLando: Fixed, and you were right to hold the line — a second review round landed on a gate that satisfies the finding without deleting the interop case I was protecting.

`packet.length < 2` -> `Malformed connack, packet too short`. Measured on this head:

```
20 00 | C0 00      Malformed connack, packet too short
20 01 00 | C0 00   Malformed connack, packet too short     (was connack{reasonCode: 0})
20 02 00 01 | C0 00   connack{reasonCode: 1} + pingreq     (v4-only server refusing a v5 CONNECT)
20 03 00 00 00 | C0 00   connack{reasonCode: 0} + pingreq
```

No `reasonCode` is fabricated any more: below 2 there is no code byte and the packet is rejected; at 2 and above the byte is real. Properties are gated at `>= 3` with the §3.2.2.3 citation, which was the FYI in your summary. `Version 5 CONNACK test 1` asserted the implied-Success behaviour, so it now asserts the error.

Why `< 2` rather than the `< 3` you asked for: remaining length 2 is the v4-format CONNACK a v4-only server sends when it refuses a v5 CONNECT, documented by `Version 4 CONACK in Version 5 mode`. A codec can be rejected-from above but not accepted-into, so MQTT.js can enforce §3.2.2 itself while this layer stays able to report the refusal. If you want the strict `< 3` anyway, say so and it is a one-line change plus that test.

test.js
robertsLando: Resolved in the parser, so this test is gone.

Remaining length 1 now asserts `Malformed connack, packet too short` rather than `rc: 0`, and the same for remaining length 0. The pipelined table keeps only the 2-byte case, annotated as the v4-format refusal. Details on the parser.js:289 thread.

robertsLando:
## Review rounds 2-7 — done, `dd17095` → `cc2a98f`

All 14 threads from the first review are resolved, and five further review rounds ran on top. **Rounds 6 and 7 came back with no blockers and no majors**, which is where I stopped.

### The theme

Every round found the same shape of bug: the fix converted *over-read* into *stop at the boundary*, but each stop returned a sentinel (`-1`, `null`, `false`) into a caller that never tested it. **Over-read became silent-accept.** All the work was at the signalling layer, not at the individual call sites.

### What each round closed

| Round | Found | Landed |
|---|---|---|
| 2 | 1 blocker, 6 majors | `_overruns` as the single boundary check, replacing seven spellings. Truncated property-length varints and property values error instead of yielding `{}` / `null`. `_parsePropertiesInto` used at all 11 call sites. A CONNACK with no reason-code byte is rejected instead of reporting Success. |
| 3 | 1 blocker, 6 majors | A packet has **two** nested bounds; `_overruns` only knew the outer one, so a property value could read past its declared Property Length into the packet's own body — `82 0d 0001 01 01 FF 0005 612f622f63 00` emitted a *valid* SUBSCRIBE. `_blockEnd` fixes it. Also: one packet could emit two `error` events; and all 11 property guards were surviving mutation testing at every site. |
| 4 | 1 major | Empty SUBSCRIBE / SUBACK payloads were accepted silently. `err.code` / `err.cmd` added. `_parseDisconnect` and `_parseAuth` extracted — they were byte-identical apart from the code table, which is why the bug in the title had to be fixed twice. |
| 5 | 1 major | `err.code` was asserted at 1 of 96 error sites; now all of them. |
| 6, 7 | none | Minors only — documentation accuracy, `err.cmd` coverage, loop uniformity. |

### On the two CONNACK threads

I pushed back on those initially, then a later security pass found a gate better than either position: `packet.length < 2`. It removes the fabricated `reasonCode: 0` that the review objected to, and keeps the 2-byte v4-format CONNACK that `Version 4 CONACK in Version 5 mode` documents. `Version 5 CONNACK test 1` asserted the old behaviour and now asserts the error. Details are on the threads.

### Verification

- **1000 tests**, up from 680. Every new mechanism dies under mutation — the matrix was re-run after each round, not assumed.
- **Differential fuzzing:** 3135 alone-vs-pipelined divergences on `master`, **0** on this head. Round 7 ran 2.9M structured cases against the previous head with 0 divergences.
- `standard` clean, `testRandom` error rate unchanged from master (~90%, i.e. no over-rejection of valid packets).
- `generate()` → `parse()` round-trips for every packet type × v3/v4/v5 × whole / chunked / pipelined: no legal packet regressed.

### Worth a maintainer's eye before merge

- **Not semver-patch.** The PR body carries the full back-compat matrix. It knowingly reverses #101 (the implied reason code for a 1-byte v5 CONNACK): a broker emitting that form connects today and will be disconnected after this.
- **A second real fix** the original description omitted: `_resetState` did not clear `_pos`, so on `master` any mid-packet error wedges the parser permanently — the next packet's Remaining Length is computed from the stale offset and the stream never resyncs.
- Error messages changed, so `err.message` is not a stable contract. Every parser error now carries `err.code` (`'MALFORMED_PACKET'`) and `err.cmd`, typed as `ParserError` and documented in the README.

One small doc fix (a README / `.d.ts` wording mismatch, from round 7) is still local and not yet pushed.

spokodev:
Seen, and thank you — that is a lot more than I brought.

The `_resetState` / `_pos` one is the part I would not have found. I was looking at a single over-read, not at the parser being wedged permanently afterwards. Same for the nested bound: I had the outer one and stopped there.

Nothing open from my side, and the way you settled the two CONNACK threads reads right to me. Yours to merge whenever the second pair of eyes is free.

